PR #2, opened 2019-03-28
Push POC files to siren-sdk repo.

m6jones:


es6/EntityFactory.js
```diff
@@ -0,0 +1,62 @@
...
+		const entityWrapped = new entityType(entity, token, entityListener);
+		onChange(entityWrapped);
+	};
+
+	// This add the listener then calls the fetch.
+	entityListener.add(href, token, onChangeWrapped);
+}
+
+// Some times the entity doesn't exists so this allows the cleanup code to be cleaner.
+export function decompose(entity) {
```
jopnick: but why not `dispose()`, for 30% less zombie

es6/EntityFactory.js
```diff
@@ -0,0 +1,62 @@
+'use strict';
+
+import 'd2l-polymer-siren-behaviors/store/entity-store.js';
+
+// This creates and fetch a new entity. Whenever the entity changes onChange is called.
```
ctwomey1: May I suggest that we use something like [JSDocs](http://usejsdoc.org/) instead of raw comments?  Some editors can support jsdocs in their intellisense to give richer descriptions.

I think formating your comments like this is all you need
```suggestion
/** This creates and fetch a new entity. Whenever the entity changes onChange is called. */
```
Especially important if we're building an sdk for wider consumption.  Totally optional though

package.json
```diff
@@ -7,5 +7,8 @@
     "test": "echo \"Error: no test specified\" && exit 1"
```
ctwomey1: I know that you're just pushing the POC files, but perhaps stubbing some super hello world tests would keep us honest as we're developing.  The first test is always the hardest one to get into a repo.

es6/EntityFactory.js
m6jones: I totally forgot

es6/EntityFactory.js
m6jones: You are completely correct here.

package.json
m6jones: Totally agree.

For setting up a test framework do you have a suggestion to which I should follow? Everyone seems to have their own style. I want to make sure it is aligned to front-end style. 

My first thought is to copy maybe the entity store repo.

m6jones: I think maybe this should be a different PR. Keep them small?

ctwomey1: yep, that sounds good.  I'm indifferent to what we use, but it seems like we commonly use mocha+chai YMMV

PR #4, opened 2019-03-31
Test skeleton and linting repo setup

m6jones:
Has the changes from https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/3

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+test/acceptance/*
+reports
+node_modules
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,10 @@
+{
+  "extends": "brightspace/polymer-3-config",
+  "globals": {
+    "fastdom": false,
+    "Promise": false
+  },
+  "env": {
+    "es6": true
+  }
+}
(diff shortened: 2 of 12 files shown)


ctwomey1:
Do we expect the sdk to be only polymer?  If we're generating just an esmodule sdk type thing we can perhaps do some unit tests against the raw classes, then reserve polymer testing for an integration suite?

m6jones:
This is going to be primary an ES6 suite. However, it will have mixins related to polymer. It could also have support for other frameworks to make it easier to use.

Here are my thoughts about this that have been rattling inside my mind after I put up the PR:
- Testing needs to test browser support since this code will be used on all browsers.
- It be really nice to not require polymer testing for the es6 modules
- Have polymer/(other framework) testing in their own test folders

Honestly, my familiarity and experience with testing framework is not great. So to get to that point I require to do some reading and learning. This does meet the main point of testing on all browsers. At the same time, I really like doing what is correct.

I am on the fence on whether to close this and just wait till I do the required learning. Anyways, that is my current thoughts.

ctwomey1:
I think you're on the right track,  I don't think it's necessary to close this pr. These are all good jump off points for testing the polymer mixin, and later a lit mixin.  

We just may want an additional suite eventually that tests things more granular as standalone modules.

m6jones:
Sounds like a plan.

PR #11, opened 2019-05-02
Consolidate entities

ctwomey1:
My original understanding of what the sdk was going to represent, was the consolidation of siren http calls into a central location.  We got the first part right, where we have a consistent way to call siren, but I noticed today that entity declarations were still spread across repos ie https://github.com/BrightspaceHypermediaComponents/enrollments/ and https://github.com/BrightspaceHypermediaComponents/organizations/

I believe that moving the entity declarations in this repo is beneficial for a couple reasons:  
1) We can expose a standard siren implementation regardless of consumers technology stack
2) With proper documentation, this could be a stepping stone to exposing a standard way for d2l external people to consume our siren apis. (imagine something like `new BrightspaceSDK().enrollments()`)

I would love to hear any feedback on this suggestion, This pr is not meant to be final but a quick representation of what I think we could have it looking like.


src/es6/Entity.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-import { entityFactory, dispose } from './EntityFactory.js';
+import { entityFactory, dispose } from './EntityFactory.js.js';
```
svanherk: Guessing this change was accidental?

src/hypermedia-constants.js
```diff
@@ -0,0 +1,393 @@
+export const Rels = {
```
svanherk: Should we even have a constants file?  This would still allow multiple uses of the same term, and I think in theory we could only need to use these "once", in the correct entity (at least for `classes` and `actions`, maybe not `rels`)

src/es6/Entity.js
jopnick: Yes, I believe he meant `./EntityFactory.js.js.js`

ctwomey1: yeah, probably just vscode trying to auto reference change

src/hypermedia-constants.js
ctwomey1: Probably not,  I was attempting to self encapsulate things, and copy pasting from the constants repo was the quickest way.

I think you're right that we'd only have them defined once, and use relations based between entities to sort things out

edwardg8:
i just realized this PR, I agree to have them in one central place probably makes more sense at the end. But I think to have the entity in its own component makes development easier because we don't know upfront what has to be in the SDK entities until all the conversion happened. After we know what the final SDK entities should be like, we can move them to this central place, otherwise this repo becomes very unstable and needs lots of releases for anything thing new added to the Entity, (i.e. just realized need to add `userActivityUsageUrl` to it)

ctwomey1:
@edwardg8 I think many releases isn't really a problem, the aws sdk releases very frequently, and the use case isn't that much different. https://github.com/aws/aws-sdk-js/releases.

Ideally, the entities should be a full representation of every possible thing a hypermedia api is exposing, and your use cases really shouldn't be exposing things that aren't possible by anyone.  So initial development of a domain might include a few bumps of versions, but once it's settled it should be a fairly low impact.

Spreading entities into the hm-component repos also harms other consumers who might want to use the apis.  While we at D2L use hypermedia components for our web projects, third parties, and nodejs projects may not or cannot.

Eventually I want to be able to go `new BrightspaceSDK().enrollments().firstname` or `new BrightspaceSDK().alerts().alertText()` but that's really only possible if our code is central

m6jones:
I think we should get this merged. Then update the entities since this also changes the file structure to be much better.

svanherk:
Are the entities here all up-to-date?  I think Edward added to `OrganizationsEntity` a few days ago

ctwomey1:
For sure aren't, Edward said his changes weren't merged yet in standup.  I can follow up and fix them when I do the flip over.  For now since they are using an older version they should be fine with what they've got.  This is a bit of a breaking change

PR #12, opened 2019-05-10
Ie11 compatibility

ctwomey1:
Very rough fix, but good enough for now i think.  There's some edgecases where it won't work exactly like new.target, but i'm not sure that matters.  If someone goes beyond the immediate safety we've provided they will have to accept the risks

### DIFF
--- a/test/entity.html
+++ b/test/entity.html
@@ -1,0 +1,30 @@
+<!doctype html>
+
+<html>
+	<head>
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1, user-scalable=yes">
+		<script src="../../@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
+
+		<!-- For IE11 -->
+		<script src="../../lie/dist/lie.polyfill.min.js"></script>
+		<script src="../../whatwg-fetch/fetch.js"></script>
+		<script src="../../wct-browser-legacy/browser.js"></script>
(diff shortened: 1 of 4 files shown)


test/entity.html
```diff
@@ -0,0 +1,29 @@
...
+		<script src="../../@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
+
+		<!-- For IE11 -->
+		<script src="../../lie/dist/lie.polyfill.min.js"></script>
+		<script src="../../whatwg-fetch/fetch.js"></script>
+		<script src="../../wct-browser-legacy/browser.js"></script>
+	</head>
+	<body>
+
+		<script type="module">
```
ctwomey1: i realize the irony of using something not compatible with ie 11 to test it.  This test is mostly to ensure we're still getting the right behavior in other browsers.  A manual test of 

```
function Entity() {
		if (this.constructor.name === Entity.name) {
			throw new TypeError('Cannot construct Entity instances directly');
		}
}
new Entity();
```
yields ```Cannot construct Entity instances directly``` in ie11

ctwomey1: Also neat, i guess WCT transpiles the html and all so that it works in ci

ctwomey1:
the additional tests i added seem to display the same problem :(  I guess ie 11 doesn't view an extended prototype as a different calling constructor

svanherk:
That last change is also what I found (bottom post of this thread: https://stackoverflow.com/questions/25140723/constructor-name-is-undefined-in-internet-explorer)

edwardg8:
Thank you Chester ++

PR #14, opened 2019-05-13
Tests with the polymer 3 mixin

m6jones:


test/utility/OrganizationEntity.js
```diff
@@ -0,0 +1,54 @@
...
+import { Entity } from '../../es6/Entity.js';
+import { SimpleEntity } from '../../es6/SimpleEntity.js';
+
+export const classes = {
+	course: 'course',
+	active: 'active',
+	inactive: 'inactive'
+};
+
+export class OrganizationEntity extends Entity {
```
ctwomey1: we should probably just merge #11 at some point so we're testing actual code instead of artificial.  It's probably out of date by now though...

es6/EntityFactory.js
```diff
@@ -25,7 +25,7 @@ export function entityFactory(entityType, href, token, onChange) {
  * @param {Object|Null} entity Object that is of an Entity type.
  */
 export function dispose(entity) {
-	entity && entity.decompose && entity.dispose();
+	entity && entity.dispose && entity.dispose();
```
ctwomey1: don't forget to regen the docs :)

test/mixin/entity-mixin-test.js
```diff
@@ -0,0 +1,134 @@
...
+				json: () => { return Promise.resolve(whatToFetch[input]); }
+			});
+		});
+	});
+
+	afterEach(() => {
+		sandbox.restore();
+	});
+
+	describe('Testing Swap', () => {
```
ctwomey1: Doesn't have to be this pr. We should be proative in testing all the things i think.  Could save some future grief

test/utility/OrganizationEntity.js
m6jones: I was wondering where those changes were - Since this is just testing the underlying Entity mocking the rest isn't an issue. Probably a name change would be better.

test/mixin/entity-mixin-test.js
m6jones: I totally agree. These tests were well over due and could have saved grief. I need more time in a day.

PR #16, opened 2019-05-16
Sequence Entity

m6jones:
Added a way to keep track of multiple sub-entities changing.


m6jones:
So even though the other PR hasn't merged yet I decided to make new entities here. This one is the sequence one. Since there is multiple activities, I made a way to track them when calling the `onChange` method by using an index.

ctwomey1:
What's the use case for the index?  It looks like it's just the array index in the subentities array, which seems odd to track to me.

m6jones:
It will allow me to keep track of the different elements based on the order they come back from the HM response.
The use case will allow me to write:
```
onSequenceActivityChange((entity) => {
         this._orgHref[entity.index] = entity.organizationHref();
}
```

Writing this out maybe I should remove that and make sure we always have `self` which would probably be safer.

es6/Entity.js
```diff
@@ -20,11 +20,23 @@ export class Entity {
...
+	/**
+	 * Get the url assigned to this entity.
+	 */
+	self() {
+		if (!this._entity || !this._entity.hasLinkByRel('self')) {
+			return;
+		}
 
+		return this._entity.getLinkByRel('self').href;
+	}
```
edwardg8: I just added the same to enrollment one, but it's better with Entity, I just have to remember removing mine after pulled in your changes

es6/Entity.js
```diff
@@ -20,11 +20,23 @@ export class Entity {
 		this._token = token;
 		this._listener = listener;
 	}
-
+	/**
+	 * Checks to see if the entity has className attached.
+	 * @param {*} className Class name you want to see if the entity has attached to it.
+	 */
 	hasClass(className) {
 		return this._entity.hasClass(className);
```
edwardg8: this line is buggy, I think you need `this._entity && this._entity.hasClass(className);`, i have it fixed in my newer changes too, if you include it here, i'll remove mine.

PR #19, opened 2019-05-22
Update getting the organization hrefs

m6jones:
I finally got my hands on the implementation of this and well these changes allow me to go through these HM


ctwomey1:
Can you add some tests :)

src/es6/Entity.js
```diff
@@ -48,10 +48,25 @@ export class Entity {
 	/**
 	 * Protected: Add a listener to a subentity of this entity.
 	 * @param {*} entityType A entity class that extends this class.
-	 * @param {*} href Href of the entity to be created
+	 * @param {*} href Href or Entity of the entity to be created
 	 * @param {*} onChange callback function that accepts an {entityType} to be called when subentity changes.
 	 */
 	_subEntity(entityType, href, onChange) {
```
m6jones: New change that gives us the ability to create a sub-entity by a complete entity.

src/es6/EntityFactory.js
```diff
@@ -48,24 +49,30 @@ class EntityListener {
 
 		window.D2L.Siren.EntityStore.addListener(this._href, this._token, this._onChange).then((removeListener) => {
 			this._removeListener = removeListener;
-			window.D2L.Siren.EntityStore.fetch(href, token);
+			if (entity) {
+				window.D2L.Siren.EntityStore.update(href, token, entity);
```
m6jones: This is the magic of it. If we already have the entity update it via this method instead of fetching it again.

test/SequenceEntity/data/sequenceRoot.js
```diff
@@ -0,0 +1,290 @@
+export const sequenceRoot = {
```
m6jones: This HM response is grabbed from the LMS copied pasted and removed most of the useless information. I kept information that isn't being used now since it will probably be required in the future.

src/sequences/SequenceEntity.js
```diff
@@ -21,15 +22,15 @@ export class SequenceEntity extends Entity {
...
+			entity && this._subEntity(SequenceEntity, entity, onChange);
 		});
 	}
 
 	onSequencedActivityChange(onChange) {
 		const sequencedActivity = this._sequencedActivity();
 
 		sequencedActivity.forEach((entity) => {
-			entity && entity.href && this._subEntity(SequencedActivityEntity, entity.href, onChange);
+			entity && this._subEntity(SequencedActivityEntity, entity, onChange);
```
m6jones: This will work if the entity is a complete entity or has a link to the complete entity

test/SequenceEntity/data/sequenceRoot.js
ctwomey1: thanks, yeah that's all i was looking for.  Just a sanity check against real data

m6jones: It caught bugs!! YAY

m6jones:
Did a rebase to cleanup the history.

src/es6/EntityFactory.js
ctwomey1: D2LFetch kinda already does this though right?

The simple cache middleware will likely have that entity already cached, and in the promise chain in fetch an update is run.  Does this actually change behavior?

m6jones: This is for embedded entities which the sequence API has so basically I could re-fetch them or just update them via this command.

ctwomey1: Right, i forgot about our embedded entities.  It does make sense to update them, I wonder if we should deep equals check to ensure we aren't notifying unecessarily

m6jones: I am taking the assumption that the entity store does the deep check. I'll check.

m6jones: It doesn't seem to :( https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-store.js#L203

m6jones: Thinking about it I think polymer is smart enough to not do any updates. It will go through some of the code though. 

This can be part of another PR.

docs/es6_Entity.js.html
```diff
@@ -76,10 +76,25 @@ <h1 class="page-title">Source: es6/Entity.js</h1>
 	/**
 	 * Protected: Add a listener to a subentity of this entity.
 	 * @param {*} entityType A entity class that extends this class.
-	 * @param {*} href Href of the entity to be created
+	 * @param {*} href Href or Entity of the entity to be created
 	 * @param {*} onChange callback function that accepts an {entityType} to be called when subentity changes.
 	 */
 	_subEntity(entityType, href, onChange) {
+		// Clean up if that href has already been added.
```
edwardg8: `href` is confusing term here at it can mean both href and entity, hard to come up with good name too. `hrefOrEntity` maybe?

m6jones: Yes, I am open to ideas. I also couldn't think of one. Maybe `hrefOrEntity`

m6jones: Maybe this is pointing out a design flaw and I should only have this as a passthrough for the `_subEntityByHref` method?

edwardg8: I put that name in the comment then think for a minute, I crossed it out, hmm maybe `source`?

edwardg8: or something more general?

docs/es6_EntityFactory.js.html
```diff
@@ -76,25 +77,31 @@ <h1 class="page-title">Source: es6/EntityFactory.js</h1>
 
 		window.D2L.Siren.EntityStore.addListener(this._href, this._token, this._onChange).then((removeListener) => {
 			this._removeListener = removeListener;
-			window.D2L.Siren.EntityStore.fetch(href, token);
+			if (entity) {
+				window.D2L.Siren.EntityStore.update(href, token, entity);
+			} else {
+				window.D2L.Siren.EntityStore.fetch(href, token);
+			}
```
edwardg8: I'm trying to understand difference between update and fetch here?

src/es6/Entity.js
edwardg8: sorry was commenting at docs, and the source code is here.

docs/es6_EntityFactory.js.html
edwardg8: Saw your comment below, sorry i'm reading docs instead of source code

m6jones: So update takes in a siren entity that has already been fetched. So it bypasses calling fetch and inserts it into the store and notifies everyone.

The reason I am doing this is because the sequence domain has a bunch of embedded HM entities.

test/index.html
```diff
@@ -22,7 +21,8 @@
 			'./skeleton-test.html',
 			'./entity.html',
 			'./mixin/entity-mixin-test.html',
-			'./EnrollmentEntity/EnrollmentEntity.html'
+			'./EnrollmentEntity/EnrollmentEntity.html',
+			'./SequenceEntity/SequenceEntity.html'
```
ciss1995: Just a heads up that we may have a conflict here. https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/20

src/sequences/SequenceEntity.js
edwardg8: what do you mean by complete entity here?

src/es6/Entity.js
```diff
@@ -48,17 +48,57 @@ export class Entity {
...
+	_subEntityByEntity(entityType, entity, onChange) {
+		if (entity.href) {
+			return this._subEntityByHref(entityType, entity.href, onChange);
+		} else if (!entity || !entity.hasLinkByRel('self')) {
+			return;
+		}
+
+		const href = entity.getLinkByRel('self').href;
+
 		// Clean up if that href has already been added.
```
svanherk: Should we just call `_subEntityByHref` here?

docs/es6_Entity.js.html
ctwomey1: It might be better to reference it more specifically.  Either embedded subentity, or linked subentity as defined by siren.

src/es6/EntityFactory.js
ctwomey1: Definitely can be another pr.  Could be backlogged in the trello board you setup.

We shouldn't think of this as being only used as polymer though.  Ideally this could be used by react, vue, lit, vanilla js, or even node.  Anything we do should adhere to generic javascript principles, not just the implementation of polymer.

m6jones: I agree!

src/es6/Entity.js
m6jones: Line 62 is different. Where it add the already fetched entity into the mix.

svanherk: Gotcha.  Could probably still do some combining around sending in the `entity` or `null`, but may be overkill

PR #20, opened 2019-05-22
US107060 Added two new entities

ciss1995:


m6jones:
Approved after lint fixes

ctwomey1:
Would be swell to get some tests added :)

ciss1995:
@ctwomey1 Ok to add tests in another PR?

ctwomey1:
I think test coverage could save us in the long run, but w/e you feel is appropriate

m6jones:
Adding tests isn't too bad. Sample tests for the enrollment entity.

ciss1995:
Tests should stop failing for IE11 After this https://github.com/Brightspace/node-siren-parser/pull/26

src/userSettings/UserSettingsEntity.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+import { Entity } from '../es6/Entity.js';
+import { Rels } from '../hypermedia-constants';
+import { Actions } from 'd2l-hypermedia-constants';
```
svanherk: Can use the Actions from the local `hypermedia-constants` file

package.json
```diff
@@ -33,6 +33,7 @@
     "lie": "^3.3.0",
     "polymer-cli": "^1.9.1",
     "wct-browser-legacy": "^1.0.1",
-    "whatwg-fetch": "^2.0.0"
+    "whatwg-fetch": "^2.0.0",
+    "d2l-hypermedia-constants": "^6"
```
svanherk: Can remove this if we use the local file

PR #23, opened 2019-05-23
US107060 Add new a function to enrollmentCollection

ciss1995:


src/enrollments/EnrollmentCollectionEntity.js
```diff
@@ -16,4 +16,12 @@ export class EnrollmentCollectionEntity extends Entity {
 		const enrollmentEntities = this._entity.getSubEntitiesByRel(Rels.userEnrollment);
 		return enrollmentEntities.map(e => e.href).filter(href => href);
 	}
+
+	getActionByName(url) {
```
ciss1995: I am using an argument here because in https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-behavior.js#L84 `getActionByName` takes `Actions.enrollments.searchMyEnrollments` or `Actions.enrollments.searchMyPinnedEnrollments`

src/enrollments/EnrollmentCollectionEntity.js
```diff
@@ -16,4 +16,12 @@ export class EnrollmentCollectionEntity extends Entity {
...
+		if (!this._entity || !this._entity.hasActionByName(url)) {
```
edwardg8: maybe use actionName instead of url?

src/enrollments/EnrollmentCollectionEntity.js
edwardg8: Better to have `searchMyEnrollmentsAction` and `searchMyPinnedEnrollmentsAction` in the SDK, so we can remove `Actions` from my courses later

src/enrollments/EnrollmentCollectionEntity.js
ciss1995: actionName could also make sense.

edwardg8: yea, both works, i'm not sure which is better now

edwardg8:
Slightly better way is to have `getActionByName` in `Entity` and have specific action name in this EnrollmentCollectionEntity I think, we want to change the parameter name `url` to actionName though

src/enrollments/EnrollmentCollectionEntity.js
m6jones: Why did we get away from this the above?

ciss1995: @m6jones I am not sure I understand what you mean?

svanherk: Yeah I agree, I think what we eventually want is that `d2l-hypermedia-constants` will mostly go away.  Each `Entity` will have the strings of the actions that make sense to it, and this will be the only place they exist (consumers like my-courses don't need to know the action strings, they just call a function).  That way, if the API or action name ever changed, we'd update the SDK and all the other components get it for free

src/enrollments/EnrollmentCollectionEntity.js
```diff
@@ -2,6 +2,7 @@
 
 import { Entity } from '../es6/Entity.js';
 import { Rels } from '../hypermedia-constants';
+import { Actions } from 'd2l-hypermedia-constants';
```
svanherk: Oh sorry, wasn't very clear.  I meant from the `hypermedia-constants` file in this repo, just changing the line above to `import { Actions, Rels } from '../hypermedia-constants';`

ciss1995: Oh okay! I didn't even see that.

PR #28, opened 2019-05-28
Egao/update enrollment

edwardg8:


edwardg8:
Moved logics from latest `enrollments` to `sdk`

edwardg8:
Will create some tests later

PR #30, opened 2019-05-30
Consortium entities

ctwomey1:
Simple setup of the consortium entities subject to change still.

Docs will be generated before merge to make the diff easier to read :)


src/consortium/ConsortiumTokenEntity.js
```diff
@@ -0,0 +1,43 @@
...
+	}
+	/**
+	 * A constructed {@link root} object
+	 * @param function onChange Callback function to be called when entity changes
+	 */
+	rootOrganizationEntity(onChange) {
+		if (!this._rootOranizationLink) {
+			return;
+		}
+		entityFactory(root, this._rootOranizationLink, this.consortiumToken(), onChange);
```
m6jones: This is a method: `this._rootOranizationLink` so it should be `this._rootOranizationLink()`

ctwomey1: oh good catch

PR #38, opened 2019-06-05
Add onEnrollmentEntityChange

ciss1995:


src/enrollments/EnrollmentCollectionEntity.js
```diff
@@ -44,4 +44,8 @@ export class EnrollmentCollectionEntity extends Entity {
 	getSearchEnrollmentsActions() {
 		return this._entity.getActionByName(Actions.enrollments.searchMyEnrollments);
 	}
+
+	onEnrollmentEntityChange(url, OnChange) {
```
ctwomey1: Would you not want to know the change for all of the subentities?  I'm not sure i understand the use case

ctwomey1: Additionally, this seems like it's registring for an onchange that may not be on the enrollmentEntityCollection.  which i'm not sure we want to do.

If you need to watch an individual enrollment Entity you might be better suited watching it as a full entity

ctwomey1: i'd still like to understand this @ciss1995 despite it getting merged

m6jones: I'd like to know too. Totally missed this but we shouldn't be adding the `url` parameter. It shouldn't require it.

PR #41, opened 2019-06-11
index sequence responses

m6jones:
Order of these sub-entities matter. 
Add the ordering to sub-entities responses.

See use: https://github.com/BrightspaceHypermediaComponents/sequences/pull/139

### DIFF
--- a/src/sequences/SequenceEntity.js
+++ b/src/sequences/SequenceEntity.js
@@ -18,6 +18,11 @@ export class SequenceEntity extends Entity {
 		return this._entity && this._entity.properties && this._entity.properties.title;
 	}
 
+	sequenceViewerApplicationHref() {
+		return this._entity && this._entity.hasLinkByRel('https://sequences.api.brightspace.com/rels/sequence-viewer-application') &&
+		this._entity.getLinkByRel('https://sequences.api.brightspace.com/rels/sequence-viewer-application').href;
+	}
+
 	onSubSequencesChange(onChange) {
 		const subSequences = this._subSequences();
 
--- a/src/sequences/SequenceEntity.js
+++ b/src/sequences/SequenceEntity.js
@@ -19,8 +19,9 @@ export class SequenceEntity extends Entity {
 	}
 
 	sequenceViewerApplicationHref() {
-		return this._entity && this._entity.hasLinkByRel('https://sequences.api.brightspace.com/rels/sequence-viewer-application') &&
-		this._entity.getLinkByRel('https://sequences.api.brightspace.com/rels/sequence-viewer-application').href;
+		const sequenceViewerRel = 'https://sequences.api.brightspace.com/rels/sequence-viewer-application';
+		return this._entity && this._entity.hasLinkByRel(sequenceViewerRel) &&
+		this._entity.getLinkByRel(sequenceViewerRel).href;
 	}
 
 	onSubSequencesChange(onChange) {
--- a/src/sequences/SequenceEntity.js
+++ b/src/sequences/SequenceEntity.js
@@ -22,11 +22,11 @@ export class SequenceEntity extends Entity {
 		const subSequences = this._subSequences();
 
 		subSequences.forEach((entity, index) => {
-			onChange = (subSequence) => {
+			const onChangeWithIndex = (subSequence) => {
 				subSequence.index = () => index;
 				onChange(subSequence);
 			};
-			entity && this._subEntity(SequenceEntity, entity, onChange);
+			entity && this._subEntity(SequenceEntity, entity, onChangeWithIndex);
 		});
 	}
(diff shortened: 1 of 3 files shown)


src/sequences/SequenceEntity.js
```diff
@@ -18,19 +18,32 @@ export class SequenceEntity extends Entity {
 		return this._entity && this._entity.properties && this._entity.properties.title;
 	}
 
+	sequenceViewerApplicationHref() {
+		return this._entity && this._entity.hasLinkByRel('https://sequences.api.brightspace.com/rels/sequence-viewer-application') &&
```
LilyAtD2L: Can we use constant for these links? At least, I think the hostname should be a constant.

src/sequences/SequenceEntity.js
```diff
@@ -18,19 +18,33 @@ export class SequenceEntity extends Entity {
...
+		this._entity.getLinkByRel(sequenceViewerRel).href;
+	}
+
 	onSubSequencesChange(onChange) {
 		const subSequences = this._subSequences();
 
-		subSequences.forEach((entity) => {
-			entity && this._subEntity(SequenceEntity, entity, onChange);
+		subSequences.forEach((entity, index) => {
+			const onChangeWithIndex = (subSequence) => {
```
ctwomey1: this feels like a not hypermedia way to do this.

Do you have a sample json?  I'd expect the entities to each have links to next and previous entities, which you could then use to create your sequence (simliar to a linked list i guess?)

m6jones: I do, I'll send it. I don't see the next/previous at this level.

test/SequenceEntity/SequenceEntity.js
```diff
@@ -13,14 +13,14 @@ describe('SequenceEntity', () => {
...
 				});
 			}, sirenSequenceRoot);
 
 			setTimeout(() => {
-				expect(orgHref['/activity/1']).to.equal('/organization/1');
-				expect(orgHref['/activity/2']).to.equal('/organization/2');
+				expect(orgHref[0]).to.equal('/organization/1');
+				expect(orgHref[1]).to.equal('/organization/2');
 				done();
 			}, 50);
```
jkehres: Is this because the parsing/loading occurs asynchronously? It would be nice to not rely on a timer in any tests. If the timer is too long, it unnecessarily slows down the tests. If the timer is too long, then the test is flaky. Is their some kind of "finished" event we could listen for instead to know that asynchronous work is complete?

Maybe something to consider for a future PR. I'm not suggesting fixing this now.

m6jones: Yes, I would love that. I don't know how though :(

test/SequenceEntity/SequenceEntity.js
```diff
@@ -13,14 +13,14 @@ describe('SequenceEntity', () => {
 			entityFactory(SequenceEntity, '/sequenceRoot', 'whatever', (entity) => {
 				entity.onSubSequencesChange((subSequence) => {
```
jkehres: Add test for `index()` here too?

src/sequences/SequenceEntity.js
```diff
@@ -18,19 +18,33 @@ export class SequenceEntity extends Entity {
 		return this._entity && this._entity.properties && this._entity.properties.title;
 	}
 
+	sequenceViewerApplicationHref() {
+		const sequenceViewerRel = 'https://sequences.api.brightspace.com/rels/sequence-viewer-application';
+		return this._entity && this._entity.hasLinkByRel(sequenceViewerRel) &&
+		this._entity.getLinkByRel(sequenceViewerRel).href;
+	}
```
jkehres: A test for this?

test/SequenceEntity/SequenceEntity.js
ctwomey1: In other repos I've wrapped the callback in a promise and resolved after the expected number of invocations. It works alright if you pair it with async/await

m6jones: Jeff proposed a solution to make the entity factory to become an async so we can then wait on these results. I believe this change is going to be pretty good.

What repos did you do your method so I can look at it?

ctwomey1: Can you resolve async things multiple times? AFAIK awaitable things can only resolve once, which messes with the pattern we've set up for a listener

m6jones: Yes, so Jeff proposed change won't mess up our pattern. From what I gather. He going to put up a PR so we can review and see it.

ctwomey1: cool, looking forward to it

PR #44, opened 2019-06-12
Let us get completion data! (From Sequences!)

m6jones:
```
class: ["completion", "incomplete"]
properties: {completed: 0, total: 1, optionalViewed: 0, optionalTotal: 5}
rel: ["item", "https://api.brightspace.com/rels/completion"]
```
This is the sub-entity that I am reading.

### DIFF
--- a/src/sequences/SequenceEntity.js
+++ b/src/sequences/SequenceEntity.js
@@ -18,15 +18,13 @@ export class SequenceEntity extends Entity {
 	 * @return An object with indices: completed, total, optionalViewed, optionalTotal, isCompleted
 	 */
 	completion() {
-		const subEntities = this._entity && this._entity.getSubEntitiesByRel && this._entity.getSubEntitiesByRel(Rels.completion);
-		if (!subEntities) {
+		const completionEntity = this._entity && this._entity.getSubEntityByRel && this._entity.getSubEntityByRel(Rels.completion);
+		if (!completionEntity) {
 			return {};
 		}
 
-		const subEntity = subEntities[0];
-
-		const completionProperties = subEntity.properties || {};
-		completionProperties.isCompleted = subEntity.hasClass && subEntity.hasClass(Classes.activities.complete);
+		const completionProperties = completionEntity.properties || {};
+		completionProperties.isCompleted = completionEntity.hasClass && completionEntity.hasClass(Classes.activities.complete);
 
 		return completionProperties;
 	}
--- a/src/sequences/SequenceEntity.js
+++ b/src/sequences/SequenceEntity.js
@@ -19,11 +19,11 @@ export class SequenceEntity extends Entity {
 	 */
 	completion() {
 		const subEntities = this._entity && this._entity.getSubEntitiesByRel && this._entity.getSubEntitiesByRel(Rels.completion);
-		const subEntity = subEntities && subEntities[0];
-
-		if (!subEntity) {
+		if (!subEntities) {
 			return {};
 		}
+
+		const subEntity = subEntities[0];
 
 		const completionProperties = subEntity.properties || {};
 		completionProperties.isCompleted = subEntity.hasClass && subEntity.hasClass(Classes.activities.complete);
--- a/src/sequences/SequenceEntity.js
+++ b/src/sequences/SequenceEntity.js
@@ -1,7 +1,7 @@
 'use strict';
 
 import { Entity } from '../es6/Entity.js';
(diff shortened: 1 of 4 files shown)


src/sequences/SequenceEntity.js
```diff
@@ -14,6 +14,23 @@ export const sequencedActivityClasses = {
...
+	completion() {
+		const subEntities = this._entity && this._entity.getSubEntitiesByRel && this._entity.getSubEntitiesByRel(Rels.completion);
+		const subEntity = subEntities.pop();
+
+		if (!subEntity) {
+			return;
+		}
+
+		const completionProperties = subEntity.properties || {};
+		completionProperties.isCompleted = !!(subEntity.hasClass && subEntity.hasClass(Classes.activities.complete));
```
AKobets: what `subEntity.hasClass` returns, function ?

test/SequenceEntity/SequenceEntity.js
```diff
@@ -59,6 +59,23 @@ describe('SequenceEntity', () => {
 			}, 50);
 		});
 
+		it('Check completion', done => {
+			const sirenSequenceRoot =  window.D2L.Hypermedia.Siren.Parse(sequenceRootMultipleTopLevel);
```
AKobets: Nitpick: I think that data should be private to the test, if somebody will change object `sequenceRootMultipleTopLevel`  in `test/SequenceEntity/data/sequenceRoot.js` it may break the test

src/sequences/SequenceEntity.js
m6jones: functions are `truthy` so it would then run the method.

AKobets: nevermind for my comment above^, line 28 explains it.

test/SequenceEntity/SequenceEntity.js
m6jones: I broke it out into its' own folder by the `json` is very long and I didn't want to clutter the test file.

src/sequences/SequenceEntity.js
```diff
@@ -14,6 +14,23 @@ export const sequencedActivityClasses = {
 };
 
 export class SequenceEntity extends Entity {
+	/**
+	 * @return An object with indices: completed, total, optionalViewed, optionalTotal, isCompleted
+	 */
+	completion() {
+		const subEntities = this._entity && this._entity.getSubEntitiesByRel && this._entity.getSubEntitiesByRel(Rels.completion);
+		const subEntity = subEntities.pop();
```
AKobets: is `subEntities` is referencing object that could be modified ( assume pop will remove item from it)

m6jones: Good point. I guess hoping it was a copy so it didn't matter. But lets be safe.

src/sequences/SequenceEntity.js
```diff
@@ -14,6 +14,23 @@ export const sequencedActivityClasses = {
...
+	 * @return An object with indices: completed, total, optionalViewed, optionalTotal, isCompleted
+	 */
+	completion() {
+		const subEntities = this._entity && this._entity.getSubEntitiesByRel && this._entity.getSubEntitiesByRel(Rels.completion);
+		const subEntity = subEntities && subEntities[0];
+
+		if (!subEntity) {
+			return {};
+		}
+
```
LilyAtD2L: nitpik: I don't think `subEntity` needs to be assigned when `subEntities` don't have substances. So, I would prefer to switch the order:
```
if (!subEnties) {
    return {};
}
const subEntity = subEntities[0];
```

m6jones: Okay

test/SequenceEntity/SequenceEntity.js
```diff
@@ -59,6 +59,23 @@ describe('SequenceEntity', () => {
...
+			const completionData = {};
+			entityFactory(SequenceEntity, '/sequenceRoot2', 'whatever', (entity) => {
+				entity.onSubSequencesChange((subSequence) => {
+					completionData[subSequence.index()] = subSequence.completion();
+				});
+			}, sirenSequenceRoot);
+
+			setTimeout(() => {
```
LilyAtD2L: Why do we use `setTimeout`? Isn't this going to be flaky?

m6jones: Jeff has a fix but right now it is the only way I could do it.

PR #50, opened 2019-07-02
Update to entity store

m6jones:
So, I updated the entity store code to es6 code. I also moved it. 

Is this a worth while change? So I move over tests?

Quick link to the previous store: https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-store.js

### DIFF
--- a/src/store/store.js
+++ b/src/store/store.js
@@ -1,0 +1,265 @@
+'use strict';
+import 'd2l-fetch/d2l-fetch.js';
+import SirenParse from 'siren-parser';
+import { getToken } from './token.js';
+
+class EntityStore {
+	constructor(fetch) {
+		this._listeners = new Map();
+		this._entities = new Map();
+		this._d2lfetch = fetch;
+	}
+
+	get(entityId, token) {
+		const lowerCaseEntityId = entityId.toLowerCase();
+		return this._entities.has(token.toString()) && this._entities.get(token.toString()).has(lowerCaseEntityId) && this._entities.get(token.toString()).get(lowerCaseEntityId);
+	}
+
+	addListener(entityId, token, listener) {
+		if (!entityId || !token.toString || typeof listener !== 'function') {
+			return;
+		}
+
+		const registrations = this._initContainer(this._listeners, entityId, token, new Map());
+		if (!registrations.has(listener)) {
+			registrations.set(listener, new Set());
+		}
+		registrations.get(listener).add(token);
+
+		return () => this.removeListener(entityId, token, listener);
+	}
+
+	removeListener(entityId, token, listener) {
+		if (!entityId || token.toString || typeof listener !== 'function' || !this._listeners) {
+			return;
+		}
+
+		const registrations = this._initContainer(this._listeners, entityId, token, new Map());
+		const tokenValues = registrations.get(listener);
+		if (!tokenValues) {
+			return;
+		}
+
+		// try to remove this specific registration, since a listener could be
+		// registered with multiple tokens (hopefully temporarily) despite sharing the same
+		// cache key (which is an internal detail)
+		if (!tokenValues.delete(token)) {
+			// we weren't aware of this particularly tokenValue for this listener, so
+			// assuemedly the component called removeListener directly with its latest
+			// values instead of the "removeListener" function returned by addListener.
+			// component is probably unregistering, so remove the listener entirely
+			registrations.delete(listener);
+			return;
+		}
+
+		// no registrations left for this listener, remove it from the list
+		if (tokenValues.size === 0) {
+			registrations.delete(listener);
+			return;
+		}
+	}
+
+	async fetch(entityId, token, bypassCache) {
+		this._initContainer(this._entities, entityId, token);
+		let entity = this.get(entityId, token);
+		if (entity && !bypassCache) {
+			return entity.entity;
+		}
+
+		this._setFetchingState(entityId, token);
+
+		const headers = new Headers();
+		token.value && headers.set('Authorization', 'Bearer ' + token.value);
+
+		if (bypassCache) {
+			headers.set('pragma', 'no-cache');
+			headers.set('cache-control', 'no-cache');
+		}
+
+		try {
+			const response = await this._d2lfetch.fetch(entityId, { headers });
+			if (!response.ok) {
+				throw response.status;
+			}
+			await this._handleCachePriming(token, response);
+			const json = await response.json();
+			entity = await SirenParse(json);
+
+			return this.update(entityId, token, entity);
+		} catch (err) {
+			return this.setError(entityId, token, err);
+		}
+	}
+
+	update(entityId, token, entity) {
+		const entityStatus = {
+			status: '',
+			entity
+		};
+		const lowerCaseEntityId = entityId.toLowerCase();
+		const tokenCache = token.toString();
+		this._initContainer(this._entities, entityId, token);
+		this._entities.get(tokenCache).set(lowerCaseEntityId, entityStatus);
+
+		this.notify(entityId, token);
+		return entity;
+	}
+
+	notify(entityId, token) {
+		const lowerCaseEntityId = entityId.toLowerCase();
+		const tokenCache = token.toString();
+		const listeners = (this._listeners.get(tokenCache) && this._listeners.get(tokenCache).get(lowerCaseEntityId)) || [];
+		const entity = this.get(entityId, tokenCache);
+
+		listeners.forEach((_, listener) => {
+			listener(entity.entity);
+		});
+	}
+
+	notifyError(entityId, token, error) {
+		const lowerCaseEntityId = entityId.toLowerCase();
+		const tokenCache = token.toString();
+		const listeners = (this._listeners.get(tokenCache) && this._listeners.get(tokenCache).get(lowerCaseEntityId)) || [];
+
+		listeners.forEach((_, listener) => {
+			listener(null, error);
+		});
+	}
+
+	setError(entityId, token, error) {
+		const lowerCaseEntityId = entityId.toLowerCase();
+		const tokenCache = token.toString();
+		this._entities.get(tokenCache).set(lowerCaseEntityId, {
+			status: 'error',
+			entity: null,
+			error: error,
+			request: null
+		});
+		this.notifyError(entityId, token, error);
+	}
+
+	_setFetchingState(entityId, token) {
+		const lowerCaseEntityId = entityId.toLowerCase();
+		const tokenCache = token.toString();
+		this._entities.get(tokenCache).set(lowerCaseEntityId, {
+			status: 'fetching',
+			entity: null
+		});
+	}
+
+	_initContainer(map, entityId, token, init) {
+		const lowerCaseEntityId = entityId.toLowerCase();
+		const tokenCache = token.toString();
+		if (!map.has(tokenCache)) {
+			map.set(tokenCache, new Map());
+		}
+		const entityMap = map.get(tokenCache);
+		if (init && !entityMap.has(lowerCaseEntityId)) {
+			entityMap.set(lowerCaseEntityId, init);
+		}
+		return entityMap.get(lowerCaseEntityId);
+	}
+
+	async _handleCachePriming(token, response) {
+		const linkHeaderValues = response.headers && response.headers.get('Link');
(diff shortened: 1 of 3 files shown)


src/store/store.js
```diff
@@ -0,0 +1,265 @@
...
+				}
+			}
+		}
+		return _links;
+	}
+}
+
+window.D2L = window.D2L || {};
+window.D2L.SirenSdk = window.D2L.SirenSdk || {};
+window.D2L.SirenSdk.EntityStore = window.D2L.SirenSdk.EntityStore || new EntityStore(window.d2lfetch);
```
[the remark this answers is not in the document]
m6jones: Yeah, I would like to get away from using the window globals. I wasn't sure how to make sure it was global to every component on the page. 

I'd also like to split this off at some point so the user can define where hold the data. That way we can use the `siren-sdk` on backend services too. 

Exporting the methods below this does remove the user from accessing or requiring knowledge that the store is stored on the window.

src/store/store.js
```diff
@@ -0,0 +1,265 @@
...
+
+export async function fetch(entityIds, token, bypassCache) {
```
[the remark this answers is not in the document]
m6jones: The design pattern for this is keeping the entity store class more lean and focused. So I am following a more functional and object approach. I also feel like these methods have more knowledge on how things are stored and such so splitting them from the class makes sense to me. 

For es6 modules, this method actually allows private code which is also quite the bonus. It also expressing the code what is intended to be used by the consumer.

For this specifically, I didn't think the class should care about how to resolve a token. It should only know how to work with a resolved token. This allows for the method within the class to use each other without resolving the token over and over again. This is great because resolving a token requires the method to be async which isn't always the desired behaviour. Such as removing a listener.

However, a consumer shouldn't have direct access to a resolved token since they can change. So the methods that are exported are all async and resolve the token. (Note for removing a listener the `addListener` method returns a function that removes the listener with an already resolved token. So this is how the user can take advantage of a resolved token without having access to it.)

With this approach it also allows us to make a batch call such that we only resolve the token once. This might be over optimization creep though. It does make it easier to work with multiple sub-entities though :D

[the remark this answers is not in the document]
ctwomey1: >The design pattern for this is keeping the entity store class more lean and focused. So I am following a more functional and object approach. I also feel like these methods have more knowledge on how things are stored and such so splitting them from the class makes sense to me.

Respectfully, i disagree.  I personally don't see anything wrong with have differing function signatures attached to the store.  It's not like you could use those functions without the entity store, so you could never take advantage of something like tree shaking or bundles. 

Additionally if you are sitting inside of an ide or some sort of smart text editor, the discoverability of the bulk actions only exists if you're inspecting the exports instead of the store itself which could be a disadvantage.

>For this specifically, I didn't think the class should care about how to resolve a token. It should only know how to work with a resolved token. This allows for the method within the class to use each other without resolving the token over and over again. This is great because resolving a token requires the method to be async which isn't always the desired behaviour. Such as removing a listener.

Not sure i really follow, You've already injected the getToken method into the file so using it inside the class doesn't seem like an issue.  You can mix async methods in for where it makes sense, and even provide async wrappers around methods you want to be synchronous.

src/store/token.js
```diff
@@ -0,0 +1,67 @@
...
+
+	isResolved() {
+		return !!this.value;
+	}
+
+	/**
+	 * From the static data from the JWT create a static key.
+	 * @param {*} token
+	 */
+	_praseCacheKey(token) {
```
jkehres: Should this be `parse` instead of `prase`?

m6jones: Yup that is a typo thank you! :)

src/store/store.js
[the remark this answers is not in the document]
m6jones: Your suggestion is to embed the global variables within the class and each instance of the class would just access them? Instead of global variable being hidden and is the class?

I still would like to see the store used in NodeJs applications as well. So the separation above brings us closer to that. I could separate these methods out into their own file. Since this is the only use case I figured why expose the class at this moment.

I also like importing the methods you need vs the entire class.

src/store/store.js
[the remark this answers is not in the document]
jkehres: Is the issue here that you want the entity store to be a singleton? Would there ever be a use case where you'd want multiple entity stores?

src/store/token.js
jkehres: This method does not access any state of the class so it could be static or you could move it outside the class and just use it as a standalone, non-exported function. The latter would also make it truly private. Another option, would be to just put all this code in the constructor since that's the only place this method is called from and the constructor is not doing anything else really.

src/store/token.js
```diff
@@ -0,0 +1,67 @@
...
+	constructor(token) {
+		this._cacheKey = this._praseCacheKey(token);
+		this._value = token;
+	}
+
+	get value() {
+		return this._value;
+	}
+
+	toString() {
```
jkehres: How often is this used? Is this just for debugging?

src/store/token.js
jkehres: Is this function just parsing a JWT? If so, I'm sure there is an existing npm library we could use for this.

src/store/store.js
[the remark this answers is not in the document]
m6jones: I figured I'd make the class not required to be a singleton. For the use case on the UI we will want it to be pretty much a singleton.

I can wrap this in a factory. Didn't really know when to call that in the UI though. I assuming it would be called in EntityFactory but then we will just be calling it on that input.

src/store/token.js
m6jones: So this method is called if you use this object as an index. Allows you to do `store = { token: test }` 

Then I used maps...

m6jones: This is also useful since it is a unique string that is the same for each token from this user.

src/store/token.js
```diff
@@ -0,0 +1,67 @@
...
+export async function getToken(token) {
+	const tokenValue = await ((typeof (token) === 'function')
+		? token()
+		: token);
+
+	return await new Token(tokenValue);
+}
+
+export function compareTokens(a, b) {
+	return a.toString() === b.toString();
```
jkehres: What's the use case for comparing tokens?

m6jones: To see if a new token is just a refreshed version of an old token.

src/store/token.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const TOKEN_COOKIE = -1;
+export const TOKEN_COOKIE_CACHE_KEY = 'cookie';
```
jkehres: How are these related to `Token` class?

src/store/token.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+export async function getToken(token) {
+	const tokenValue = await ((typeof (token) === 'function')
+		? token()
+		: token);
+
+	return await new Token(tokenValue);
```
jkehres: `new Token(...)` doesn't return a promise so `await` is unnecessary.

src/store/token.js
m6jones: If there is no token then the token will be sent through a cookie. So it sets the state of token to cookies deal with it.

src/store/token.js
m6jones: It is taking our token then only parsing the data that doesn't change between token refreshes. I don't think using a parser would add anything. It wouldn't replace this method.

jkehres: I agree that a parser would not functionally add anything but it's code that someone else writes/maintains/tests so less work for us. I don't currently see a test for this code so that's additional work you'll eventually need to do.

src/store/token.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+export async function getToken(token) {
```
jkehres: You could model this after `Array.from()` and make this a static `from()` method on the `Token` class (e.g. `Token.from()`);

src/store/token.js
```diff
@@ -0,0 +1,67 @@
...
+	const tokenValue = await ((typeof (token) === 'function')
+		? token()
+		: token);
+
+	return await new Token(tokenValue);
+}
+
+export function compareTokens(a, b) {
```
jkehres: Using the word "compare" makes me think of a comparison function that returns an integer (e.g. 0, < 0, > 0) like what would be used with the `sort()` method of an array. However, this function returns a boolean since it's just an equality check perhaps use the term "equals" instead of "compare". Perhaps model this after Java's `Object.equals()` method and make this a member method of the `Token` class.

src/store/token.js
```diff
@@ -0,0 +1,67 @@
...
+
+	get value() {
+		return this._value;
+	}
+
+	toString() {
+		return this._cacheKey;
+	}
+
+	isResolved() {
```
jkehres: What does a resolved token mean?

src/store/token.js
jkehres: Do these need to be exported or are they only used within the `Token` class?

jkehres:
Once you finalize the design of this code. You should add some tests to this PR.

src/store/token.js
jkehres: This would mean you'd need to export the `Token` class, though, which you might not want as you can't make the constructor private and enforce the use of this factory function.

src/store/store.js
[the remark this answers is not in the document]
jkehres: You could use lazy initialization to create the singleton (e.g. create it the first time someone tries to access it). Alternatively, you could instantiate your singleton in the global scope of this file. This means the singleton gets created when the file is first loaded. Assuming `import` works similar to `require`, then there is caching so a file only gets loaded once even if it's imported multiple places. This would guard against multiple singletons being created.

src/store/store.js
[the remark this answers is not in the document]
jkehres: > I also like importing the methods you need vs the entire class.

Is that just because you can't have private members in a class so you're using module exports to achieve the same results?

src/store/token.js
jkehres: Alternatively, you could make getting a token the problem of the caller of this module (i.e. they should already have a token in hand so if they've got a function instead of a value it's their responsibility to invoke the function). This would allow this file to be solely about the `Token` class.

src/store/store.js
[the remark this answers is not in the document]
jkehres: For example:
```
class EntityStore { ... }
export const singleton = new EntityStore( ... );
```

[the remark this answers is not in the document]
ctwomey1: I'd tread carefully assuming that the module loaders are uniform.  Whatwg defines the module map as such https://html.spec.whatwg.org/multipage/webappapis.html#module-map, but you'd have to also ensure that the transpiling to es5 also behaves the same.

It's possible that we don't care about it though, but it would feel slightly safer to factory method with a cache on windoworglobal

src/store/token.js
m6jones: Change of name I agree. Compare doesn't make sense.

I thought object.equals wouldn't be value based but it is! so I can change it to that.

src/store/token.js
m6jones: That it has been processed with no errors.

m6jones: Is this token a valid instance of `Token`.

src/store/store.js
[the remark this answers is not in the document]
m6jones: So making a factory sounds like the best way. So I'll do that for sure.

src/store/store.js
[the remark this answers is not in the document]
m6jones: This may be the underline reason since I really like a clean interface. Here the class is implementation detail that doesn't need to be exposed. The store only requires the methods we are exporting and the class just makes implementation of those easier.

Exporting the class here seems over kill and making the interface complicated. This will then make it harder to maintain and so on.

src/store/token.js
m6jones: There are tests in the other repo for this. I'll be bringing it over after I have the architecture sorted.

src/store/token.js
m6jones: Good point, I don't think they do need to be exported.

src/store/token.js
m6jones: The person with the token, won't know what type of token they will have and it would be a mess if everyone had to test for it.

That is why I kept it outside the class so I could export it and control the interface.

PR #56, opened 2019-07-23
Add a StateTree helper to simplify state change updates for nested onChange handlers

AllanKerr:
# Changes
This pull request adds a `StateTree` class that can be used when dealing with nested `onChange` handlers to ensure that all updates are generated every time any part of the tree changes.


m6jones:
In this repo we use java docs. Mind adding some? Then you can `npm run generate:docs`

src/helpers/StateTree.js
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * StateTree helper class to simplify receiving state changed updates for nested onChange handlers.
+ */
+export class StateTree {
+	/**
+	 * Creates a new StateTree.
+	 * @param {Function} onStateChanged A state change handler that is called whenever the state of the tree changes by adding, replacing, or removing items.
+	 * @param {*} [item] The item to be stored at the root of the tree.
+	 */
+	constructor(onStateChanged, item, root) {
```
ctwomey1: Parameter root should be documented

src/helpers/StateTree.js
```diff
@@ -0,0 +1,62 @@
...
+			items = items.concat(child.items());
+		});
+		return items;
+	}
+
+	_notifyStateChanged(hasStateChanged) {
+		if (!hasStateChanged) {
+			return;
+		}
+		this._onStateChanged(this._root);
```
ctwomey1: What's your use case that your root entity cares about children changing?  My assumption is embedded representation siren entities.  If each child has a call back available for updating your ui state, updating the root would cause a fuller reflow of all the callbacks right?

Like a typical use case for many nested children ends up looking like:

```
consotriumTokenCollection.consortiumTokenEntities((consortiumEntity) => {
	consortiumEntity.rootOrganizationEntity((rootEntity) => {
         	rootEntity.organization((orgEntity) => {
                        //handler stuff that would get reinited
                }
         }
}
```

src/helpers/StateTree.js
```diff
@@ -0,0 +1,62 @@
...
+		this._notifyStateChanged(hasStateChanged);
+	}
+
+	/**
+	 * Retrieves all items stored in the tree without guaranteeing any particular order.
+	 * @returns {Array} All items stored in the tree.
+	 */
+	items() {
+		let items = this._item ? [this._item] : [];
+		this._children.forEach(child => {
```
ctwomey1: Given the JSDoc, shouldn't this be this._root._children if you want all the children?

This assumes that you are always calling items at the root, but you might be on a child atm.  Could either update the jsdoc to be something like 

```
Retrieves all children from this node stored in the tree without guaranteeing any partiular order
```

or change it to go from the root.

m6jones:
Here is the use case: https://github.com/BrightspaceHypermediaComponents/enrollments/pull/144

This will be useful in other repos as well so that is why I suggested we bring it over here as a helper.

Basically, this helps manage the chains the sdk is creating especially when using sequence domain. Since each step of the chain we need to clean up properly. So this data structure manages the details for us.

Some of the giant chains that are being created are screaming there is a design flaw some where so this is a stop gap solution until we figure out how to fix that.

ctwomey1:
I'd agree that there's a definite flaw.  We could either lean into a more flux model where we simply fire actions then listen to the results, or perhaps jump into observables https://github.com/tc39/proposal-observable where we observe the entire sdk, but filter down to events we care about, and then just unsubscribe when we detach from the sdk.  I'll create an issue where we can outline the problem and discuss the solutions outside of this pr.

Anyways https://github.com/BrightspaceHypermediaComponents/enrollments/pull/144/files#diff-bc6e475d126bd0aca44f8e5152cde3f8R404-R422

Looking at that above link, i think that you can make the helper a "private" class.  Managing the state shouldn't be delegated to consumers of the sdk, because it's super easy to mess up or forget to track children, and if we want to change how we manage state it's much easier if it's not distributed everywhere.

m6jones:
Do you have ideas on how to make it private? Right now, this class make it easier to hold the state within respect to a webcomponent. Not sure how to make it general such that it can be private.

ctwomey1:
I mean it's "Private" by just not using it outside of the sdk.  We can't really stop people from using it.

If i was to move it in side the sdk, i'd probably tie it into the core concept of an Entity.  In this case, organization entity could have the concept of it's children, and each child would get disposed when the organization is changed.  Granted I haven't sat down and fully thought it out, it could be harder than i am inferring

m6jones:
Yeah, so the children entities do get disposed. The issue is the web-component save data related to these entities and that also needs to be reverted.

ctwomey1:
We could introduce a callback for when things have been disposed or something.  Ideally we wouldn't expect a caller to track state but react to it instead.  I think observables could be a good option

m6jones:
Yeah, lets look into better options but this at least will help us somewhat keep track of what will need updates when we do figure out the better solution.

PR #58, opened 2019-07-30
DE35341-fix my courses pinning defect

edwardg8:


src/es6/EntityFactory.js
```diff
@@ -10,15 +10,15 @@ import 'd2l-polymer-siren-behaviors/store/entity-store.js';
  * @param {Function} onChange Callback function that accepts an {entityType} to be called when entity changes.
  * @param {Object} entity (Optional) Entity that has already been fetched.
  */
-export function entityFactory(entityType, href, token, onChange, entity) {
+export async function entityFactory(entityType, href, token, onChange, entity) {
```
edwardg8: i'm not too sure why, but this fixes the pinning defect DE35341

m6jones:
Is this PR suppose to go into the new store code? Or was there a pinning issue in general?

edwardg8:
Stacey, Cheick and I had some discussions around solutions for the two defects yesterday. We'll likely be doing some hot fixes, so we are trying to find the minimum changes from your PR that may fix the issues. The first thing we tried is to convert the `EntityFactory` to async function, it fixes the issue right away without breaking pinning, so there is some race condition going on there we think. We are not sure what the best fix is at this moment. Stacey has also tried to use async and await in my courses widget itself, that seems also working, so definitely race conditions when resolving the promise

m6jones:
This sort of worries me since the async I added from my PR was only added to make testing easier.

edwardg8:
yea, this is definitely a random fix, with async it might have slightly changed the execution time abit that fixes the problem. This is like shooting in the dark and hit the target, it's not a proper fix, but I don't know what can be done better.

svanherk:
I'm seeing this defect also fixed if I add this to a my-courses function, instead of the sdk.  So I think our promise chain in my-courses is broken somewhere, I'm not sure which is the best to fix it

m6jones:
The sdk doesn't really support promises yet. Since in the general use case web components should never have to wait on a response. 

So that could be a main issue.

svanherk:
Yeah talking with Edward, we don't think we should _need_ to expect a promise here at this point, the chain in my-courses is returning all the way up for no reason.  So I'm going to put up a PR shortly with an alternative fix that just touches my-courses

edwardg8:
Yea, Stacey is going to remove the unnecessary return of "promise" from use of entity factory, it wasn't needed anyway, maybe that cleans up the code and fixes the defect, that'll be much better self contained solution

edwardg8:
Sorry I just made the same change, and waited for the build to test out, and distracted by the talk, should have pushed it. Now it breaks all the tests for populateEnrollments.

PR #63, opened 2019-09-18
add placeholder entities for assignment activities

mdgbayly:
These are some placeholder entities for Assignments for use in the new assignment activity editor.
I have not fully got my head yet around how these `Entity` classes work and how they can be used to nest entities, but it looks interesting and flexible to different approaches. 

The `Entity` base class used with the `entity-mixin-lit` for the web components basically provides a wrapper interacting with the EntityStore that perhaps makes it easier to manage that interaction.

The test class `siren-sdk-organization-info-lit.js` in this repo shows an example of how these components can be used to work together.

https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/test/utility/siren-sdk-organization-info-lit.js

The new activities web components that i will push in a separate PR (in the `activities` repo) will be using these Entity components to load both the assignment activity usage siren response and the assignment folder siren response.

For now the entities just provide access to the main assignment folder API endpoint, and then to the basic assignment properties like name. But as we extend the features of the editors we will be extending these entities.

### DIFF
--- a/src/activities/assignments/AssignmentActivityEntity.js
+++ b/src/activities/assignments/AssignmentActivityEntity.js
@@ -1,0 +1,13 @@
+'use strict';
+
+import { Entity } from '../../es6/Entity.js';
+
+/**
+ * AssignmentActivityEntity class representation of a d2l AssignmentActivity.
+ */
+export class AssignmentActivityEntity extends Entity {
+	name() {
+		return this._entity && this._entity.properties && this._entity.properties.name;
+	}
+}
+
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "test:lint": "npm run test:lint:js",
     "test:lint:js": "eslint . --ext .js,.html test/**/*.js test/**/*.html demo/**/*.js demo/**/*.html",
     "test:polymer:local": "polymer test --skip-plugin sauce",
-    "test:polymer:sauce": "polymer test --skip-plugin local"
+    "test:polymer:sauce": "polymer test --skip-plugin local",
+    "serve": "polymer serve"
   },
   "author": "",
   "license": "MIT",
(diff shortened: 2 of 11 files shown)


src/activities/assignments/AssignmentActivityEntity.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+import { Entity } from '../../es6/Entity.js';
+
+/**
+ * AssignmentActivityEntity class representation of a d2l AssignmentActivity.
+ */
+export class AssignmentActivityEntity extends Entity {
```
ryantmer: I'm probably being too literal/strict here, but wouldn't this be `AssignmentEntity`, since we're outside of the `activities` HM domain at this point? (Although I recognize that all of this is kind of within the activities domain, in a way... 🤔)

mdgbayly: Yeh - I ummed and ahhed over the naming... i don't feel strongly either way... an assignment is an activity but would we also have a representation of an AssignmentActivity that is not an AssignmentActivityUsage but is also not the core Assignment. Maybe. Eric's original model had ActivityTemplate, ActivityInstance (which seems to have become ActivityUsage?), and AssignedActivity. So maybe if we had an entity that was not a usage but was also just the activity common stuff that would be an AssignmentActivityEntity.

I think I'll rename it to AssignmentEntity just because it's shorter!

ryantmer: ActivityUsage I believe is the old terminology, whereas ActivityTemplate/Instance/Assigned are the new ones, so old -> new mapping becomes

Activity -> ActivityTemplate
ActivityUsage -> ActivityInstance
UserActivityUsage -> AssignedActivity

I do like these names better tbh, sadly the codebase is full of the old ones :/

PR #64, opened 2019-09-23
US110193 add function to determine if editing name is permitted

mdulat:
Goes with https://github.com/BrightspaceHypermediaComponents/activities/pull/380

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -9,5 +9,8 @@ export class AssignmentEntity extends Entity {
 	name() {
 		return this._entity && this._entity.properties && this._entity.properties.name;
 	}
+	canEditName() {
+		return this._entity && this._entity.hasActionByName('update-name');
+	}
 }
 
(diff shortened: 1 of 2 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -1,13 +1,16 @@
...
+
+/**
+ * AssignmentEntity class representation of a d2l Assignment.
+ */
+export class AssignmentEntity extends Entity {
+	name() {
+		return this._entity && this._entity.properties && this._entity.properties.name;
+	}
+	canEditName() {
+		return this._entity && this._entity.hasActionByName('update-name');
```
mdgbayly: super complex implementation I know ;-), but i guess we could try and get into the habit of writing unit tests... of course then we write them for the easy stuff and never for the harder stuff...but...

mdgbayly:
> Looks like the line endings all changed - there is an `.editorconfig` that specifies LF for this repo, but not sure if they were correct before or now! :)

Yeh - that would be my bad... not sure how they ended up with CRLF... i think must have copied the files, and the editor config didn't correct them.

PR #65, opened 2019-09-25
US110193 add siren fetch action and ability to save assignment name

mdulat:


mdulat:
> Looks good - i think there were some unit tests that went with the legacy behavior, so maybe we can bring those over too?

Sure, I'll put into subsequent PR.

PR #66, opened 2019-09-26
US110193 add unit tests for siren actions

mdulat:


test/mixin/siren-fetch-mixin-lit-test.js
```diff
@@ -0,0 +1,179 @@
...
+			{ 'rel': ['self'], 'href': 'http://api.x.io/orders/42' },
+			{ 'rel': ['previous'], 'href': 'http://api.x.io/orders/41' },
+			{ 'rel': ['next'], 'href': 'http://api.x.io/orders/43' }
+		]
+	};
+
+	setup(function() {
+		sandbox = sinon.sandbox.create();
+		element = fixture('basic');
+		element.token = 'foozleberries';
```
ctwomey1: I think that you need to drain the action queue and purge the siren store inbetween tests.  Otherwise you can have things bleeding over.

test/mixin/siren-fetch-mixin-lit-test.js
```diff
@@ -0,0 +1,179 @@
...
+
+	suite('smoke test', function() {
+		test('can be instantiated', function() {
+			expect(element.tagName.toLowerCase()).to.equal('siren-fetch-mixin-lit-test-component');
+		});
+	});
+
+	suite('send action unit tests', function() {
+		setup(function() {
+			sinon.stub(window, 'fetch');
```
ctwomey1: Might want to stub https://github.com/Brightspace/d2l-fetch/blob/master/src/d2lfetch.js#L7 instead of the global window fetch.

That way you only trap the things you are expecting rather than everything that calls fetch on the page

test/mixin/siren-fetch-mixin-lit-test.js
```diff
@@ -0,0 +1,179 @@
...
+			);
+			var firstCall = element.performSirenAction(testAction);
+			element.performSirenAction(testAction2);
+			return firstCall;
+		});
+
+		test('immediate actions skip queue', function() {
+			fetchStub.withArgs('http://api.x.io/orders/42/items').returns(
+				new Promise(function(resolve) {
+					setTimeout(function() {
```
ctwomey1: Are you trying to alter the micro tasks that get run by doing an empty setTimeout?  Or what is this for?

test/mixin/siren-fetch-mixin-lit-test.js
```diff
@@ -0,0 +1,179 @@
...
+				})
+			);
+			fetchStub.withArgs('http://api.x.io/orders/42/items/5').returns(
+				new Promise(function(resolve) {
+					resolve(res2);
+				})
+			);
+			var firstCall = element.performSirenAction(testAction);
+			element.performSirenAction(testAction2);
+			return firstCall;
```
ctwomey1: should you have a sinon spy or something on https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/action-queue.js#L7 so that you can assert that things are being queued or not?

test/mixin/siren-fetch-mixin-lit-test.js
mdulat: The tests were copied from https://github.com/Brightspace/polymer-siren-behaviors/blob/master/test/siren-action-behavior.js but I believe the reason for the timeout is just to allow the second call to be made before checking the results. In any case, there was an inconsistency in the way the two "enqueue action unit tests" were being checked, so I made it consistent.

PR #67, opened 2019-09-26
Refactor lit entity mixin to support pending state and add PendingContainer

mdgbayly:
Refactor the lit entity mixin to send a `pending-state` event with a promise that will be resolved when the component's entity has been loaded.

Add a PendingContainer mixin that can be used to track `pending-state` events from child components and toggle a container `hasPendingChildren` property when all children are loaded.

This could be used to render a skeleton in the parent component whilst the children are loading.


src/mixin/entity-mixin-lit.js
```diff
@@ -18,60 +18,27 @@ export const EntityMixinLit = superclass => class extends superclass {
 			/**
 			 * Entity object that extends the Entity class.
 			 */
-			_entity: { type: Object },
```
mdgbayly: On the fence about this rename, but this was in response to some of Dave L's comments about not overusing _properties and _methods.

I find them pretty ugly looking and it feels like the entity could be a first class property of entity baed components which could be set directly.

But I could be convinced otherwise of this!

src/mixin/entity-mixin-lit.js
```diff
@@ -18,60 +18,27 @@ export const EntityMixinLit = superclass => class extends superclass {
...
+			entity: { type: Object }
 		};
 	}
 
-	constructor() {
-		super();
-	}
-
-	set href(href) {
```
mdgbayly: A lot of these custom property setters seemed superfluous as they are just copying the default lit implementation, so I removed them.

src/mixin/entity-mixin-lit.js
```diff
@@ -18,60 +18,27 @@ export const EntityMixinLit = superclass => class extends superclass {
...
-
 	disconnectedCallback() {
 		super.disconnectedCallback();
 		// this calls this._entity.dispose() if entity is actually an entity.
 		// Note this will dispose all child entities used.
-		dispose(this._entity);
+		dispose(this.entity);
 	}
 
-	requestUpdate(propChanges, oldValue) {
```
mdgbayly: I was on the fence about where best to put our hook to load the entity too.
I went with shouldUpdate because I wanted to prevent renders when we don't have a token or an href yet, and it seemed simpler to just keep the logic all together.

src/mixin/entity-mixin-lit.js
```diff
@@ -85,11 +52,27 @@ export const EntityMixinLit = superclass => class extends superclass {
 		}
 	}
 
-	__onHrefChange(href, token) {
-		dispose(this._entity); // Make sure the entity is cleaned up before setting a new one.
+	_getEntity() {
+		dispose(this.entity);
 		if (typeof this._entityType === 'function') {
```
mdgbayly: This is an alternative approach to using `runAsync` that still gives us `pending-state` support.
These events are used by the `PendingContainer` mixin to track how many children are pending.
This could be used to render skeleton components in the main container until all the children are ready.

test/mixin/simple-entity-mixin-test.js
```diff
@@ -0,0 +1,110 @@
+import 'd2l-fetch/d2l-fetch.js';
+import { html, fixture, expect } from '@open-wc/testing';
```
mdgbayly: Experimenting with using `@open-wc/testing` framework as it seems simpler to use than the legacy polymer stuff.

test/mixin/simple-entity-mixin-test.js
```diff
@@ -0,0 +1,110 @@
...
+				<div id="organization-name">Course Name</div>
+			`);
+	});
+
+	it('updates a name', async() => {
+		const el = await fixture(html`
+				<siren-sdk-simple-organization token="whatever"></siren-sdk-simple-organization>
+			`);
+
+		await load(el, '/organization.json');
```
mdgbayly: Trying to come up with a way to test things which so much async stuff happening. This load helper is kind of ugly and I might try and come up with a different implementation using the `@open-wc/testing` event helpers.

src/mixin/entity-mixin-lit.js
ctwomey1: Something to consider is d2lfetch-auth is almost always in the middleware chain. https://github.com/Brightspace/d2l-fetch-auth/blob/master/src/unframed/d2lfetch-auth.js.  Tokens are semi optional

dbatiste:
A somewhat minor point: so far all of our core mixin names end with "Mixin".  So instead of `entity-mixin-lit.js` and `EntityMixinLit`, we might have named this `lit-entity-mixin` and `LitEntityMixin`.  I feel like I don't have a great argument for it though, other than perhaps consistency and ease of reading.

src/mixin/entity-mixin-lit.js
dbatiste: It feels odd to me that there would be both an `href` and `entity` on the public API, like there is opportunity for them to get of sync if entity is not read-only to public.

src/mixin/entity-mixin-lit.js
```diff
@@ -85,11 +52,27 @@ export const EntityMixinLit = superclass => class extends superclass {
...
-			entityFactory(this._entityType, href, token, entity => {
-				this._entity = entity;
+			let pendingResolve;
+			var pendingPromise = new Promise(function(resolve) {
+				pendingResolve = resolve;
+			});
+
+			const pendingEvent = new CustomEvent('pending-state', {
```
dbatiste: Do we want to align on a D2L specific event, `d2l-pending-state`?

src/mixin/pending-container.js
```diff
@@ -0,0 +1,57 @@
+
+export const PendingContainer = superclass => class extends superclass {
```
dbatiste: Does it make sense for this to be a mixin?

src/mixin/pending-container.js
```diff
@@ -0,0 +1,57 @@
...
+
+	connectedCallback() {
+		this.addEventListener('pending-state', this.__onPendingState);
+		if (super.connectedCallback) {
+			super.connectedCallback();
+		}
+	}
+
+	disconnectedCallback() {
+		this.addEventListener('pending-state', this.__onPendingState);
```
dbatiste: Looks like this should be `removeEventListener`.

However, if this was rendering an element, the connectedCallback and disconnectedCallback could be avoided.

```
render() {
   return html`<div @d2l-pending-state="${...}"><slot></slot></div>`;
}
```

src/mixin/pending-container.js
dbatiste: Sorry, the name was throwing me off - this is already a mixin.  :)  I think it should be called `PendingContainerMixin`.

src/mixin/pending-container.js
```diff
@@ -0,0 +1,57 @@
+
+export const PendingContainer = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			hasPendingChildren: {
```
dbatiste: It looks like these don't need to be properties, or public.

src/mixin/entity-mixin-lit.js
mdgbayly: I was using  `pending-state` for now because that is what the `@brightspace-ui/core` `runAsync` directive is using and I thought it might be good to keep the PendingContainer usable with runAsync too.  But I'm ok to switch it to use `d2l-pending-state` if we prefer.

src/mixin/pending-container.js
mdgbayly: I hadn't initially considered using an element. Was basically just copying the PendingContainer  mixin from Justin's demos.  I'd probably like to leave it as a mixin for now whilst we try it out in the new activity editor.  If we find that everyone wants to use a helper element at the top level, then that element could extend this mixin and become the standard way for people to use it.

src/mixin/entity-mixin-lit.js
mdgbayly: OK - i'll switch it back to `_entity`.

src/mixin/pending-container.js
dbatiste: Yep I totally agree with it being a mixin, but I think it should be renamed.

src/mixin/entity-mixin-lit.js
mdgbayly: Yeh, that's a good point. I'm not clear on the best way to handle that though. As if the href is set before the token, but you really intend to set a token, it can end up making the request without the token.

src/mixin/entity-mixin-lit.js
dbatiste: Yeah, I don't feel strongly about it.  I don't think anything is using `pending-state` from `run-async` so we could update them both to be consistent.  The original `pending-state` event name was from Justin's PR which is unlikely to be merged.

src/mixin/pending-container.js
dbatiste: Since this is actually a mixin, no need to make the `render` more complex.

src/mixin/pending-container.js
```diff
@@ -0,0 +1,57 @@
...
+			this.pendingCount++;
+			try {
+				await promise;
+			} catch (e) {
+				// to supress uncaught exception logs
+			} finally {
+				this.pendingCount--;
+				if (this.pendingCount === 0) {
+					this.hasPendingChildren = false;
+					const pendingEvent = new CustomEvent('pending-resolved', {
```
dbatiste: Just thinking about how this component will be used...

Is the `pending-resolved` event just for testing purposes?  It appears (in the sample below) that re-render is triggered by a change to `hasPendingChildren`.  It looks like `pendingCount` _could_ be private, and doesn't really need to be a property.  Thoughts?

mdgbayly: Yeh the `pending-resolved` event was my cheap shot at making this more testable....

And yeh, had not given too much thought to the usage of the pending fields.

I think I misread Justin's initial Typescript and thought he had made them both properties, but looking again it looks like only the `hasPendingChildren` is a property and he has annotated it as protected, so I guess it should become:

```
static get properties() {
		return {
			_hasPendingChildren: {
				type: Boolean
			}
		};
	}
```

and as you say pendingCount can just be a private field.

test/mixin/pending-container-test.js
```diff
@@ -0,0 +1,64 @@
...
+	});
+
+	it('waits for children', async() => {
+		const el = await fixture(html`
+				<siren-sdk-pending-container href='/organization.json' token="whatever"></siren-sdk-pending-container>
+			`);
+
+		expect(el).shadowDom.to.equal(`
+				<div>Loading...</div>
+				<siren-sdk-simple-organization hidden="" href="/organization.json" token="whatever"></siren-sdk-simple-organization>
```
ryantmer: Is the fact that it's `hidden=""` and not just the boolean attribute `hidden` a quirk of testing here? Seems strange, but based on your description of your experience in working with these tests recently, wouldn't surprise me

test/mixin/simple-entity-mixin-test.js
```diff
@@ -0,0 +1,110 @@
...
+
+	el.addEventListener('pending-state', wait);
+	el.href = href;
+
+	await loading;
+}
+
+describe('Simple Entity Mixin Test', function() {
+
+	// this.timeout(180000);
```
ryantmer: What ended up being the solution for the timeouts here?

mdgbayly: Oh, I was just using that to make it easier to debug the tests. Mocha will timeout the test after 10 seconds, so if you are debugging a test it can be helpful to use a longer timeout so after debugging you can let the test run to a successful/normal completion.

test/mixin/pending-container-test.js
mdgbayly: Yeh, it seems to be a quirk of how lit-element sets these properties. When the boolean attribute is present it seems to get an empty `=""`.

src/mixin/pending-container.js
```diff
@@ -0,0 +1,57 @@
...
+	disconnectedCallback() {
+		this.addEventListener('pending-state', this.__onPendingState);
+		if (super.disconnectedCallback) {
+			super.disconnectedCallback();
+		}
+	}
+
+	async __onPendingState(e) {
+		const promise = e.detail.promise;
+		if (promise) {
```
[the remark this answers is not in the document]
mdgbayly: Yeh - will look at it - this was a straight copy of a prototype component from one of the lit guys.

src/mixin/pending-container.js
```diff
@@ -0,0 +1,57 @@
...
+
+	constructor() {
+		super();
+		this.pendingCount = 0;
+		this.hasPendingChildren = false;
+	}
+
+	connectedCallback() {
+		this.addEventListener('pending-state', this.__onPendingState);
+		if (super.connectedCallback) {
```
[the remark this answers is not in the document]
mdgbayly: Yeh , not sure ... again my defence is i copied it from the lit-element guy's (probably hacked together) code...

test/mixin/pending-container-test.js
dbatiste: When using Lit-Element's `?myattr="${this.someBool}"` directive for boolean attributes, it will either add the attribute as `myattr`, or it won't.  I think what we are seeing here may be the product of the `shadowDom` helper, which may be using `innerHTML` or `outerHTML` properties internally, which will result in `myattr=""`.  For example, inspecting the DOM in the browser console should not show these empty strings.

I think this oddity could be avoided by expecting `el.shadowRoot.querySelector('div').hasAttribute('hidden')` to be `false`, and another expectation statement for the siren-sdk element.

test/mixin/simple-entity-mixin-test.js
ryantmer: Ah okay, makes sense. Mostly I was concerned that a test was taking longer than 10 seconds, which is rather long - but I do also recall that Sauce in particular can have some serious issues with doing things quickly 😢

src/mixin/pending-container-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+		return {
+			_hasPendingChildren: {
+				type: Boolean
+			}
+		};
+	}
+
+	constructor() {
+		super();
+		this.__pendingCount = 0;
```
ryantmer: The `__` and `_` in Polymer always kind of confused me - is there a convention for this?

wct.conf.json
```diff
@@ -37,11 +37,6 @@
 					"browserName": "microsoftedge",
 					"platform": "Windows 10",
 					"version": ""
-				},
-				{
-					"browserName": "internet explorer",
```
ryantmer: 🙌 🙌 🙌

src/mixin/pending-container-mixin.js
dbatiste: We've had a couple of conversations about this but I feel like it's still a loose-end perhaps because folks were/are focused on other important things.  :)

[This StackOverflow question/answer](https://stackoverflow.com/questions/35946339/double-underscore-vs-single-underscore-in-nodejs) seems to indicate double underscore is not an "official" convention.  Apparently single underscore is the only "unofficial" convention for indicating "private" members, and the double underscore came about due to 2 NodeJS global variables, `__dirname` and `__filename`.

I think we've used double underscore in certain places in effort to avoid possible collision between components and mixins.  I think the different underscores have also been used by some to differentiate between protected and private - this muddies things up even more, and I don't think there is much convention for this.

My personal view is that we should use single underscore for private variables everywhere, and behaviors/mixins should name their members to best avoid such collision.  I'd rather have a funky verbose name in one mixin than force every consumer of that mixin to have an overly verbose name.

The "protected" case is interesting, and I am not sure how these should be named. i.e. when a mixin defines a member it intends the sub-class to use, but it does not expect it to be used by others.  I tend to think it should simply be treated like a public member, but I acknowledge that I prefer `_entity` to be "private", mostly because it is unsettling to me to have both `entity` and `href` as public members.

I am happy to use whatever convention we agree on - I'd rather be consistent than not, even if it's not my preference.  :)

mdgbayly: I see a lot of usages of `_` and `__` in some of the `brightspace-ui/core` projects. My understanding has always been that yeh `_` is protected, `__` is private, for what little value that has in JS anyway. I don't feel strongly about it either way but will probably leave as is until someone says it's categorically something we should stop doing.

mdgbayly:
Thanks for all the comments and feedback everyone. I think there are still some things that could be done to improve this, particularly around the promise/entityFactory callback handling, but for now I'm going to merge and run with it to get a bit more experience actually using it with some of the components we are building.

If there is anything about it that people still needs fixing ASAP, let me know and I'll add in a follow up PR.

src/mixin/pending-container-mixin.js
dbatiste: Yeah we haven't been consistent in our use in `core`.  I'd like to fix that, and in `core` it has usually been used to avoid collision between private mixin members and private component members with the same name.  Technically it doesn't provide any real runtime protection, but it helps understand the code since if infers scope.

src/mixin/entity-mixin-lit.js
m6jones: So these were not superfluous. They made it so updateRequests were not called. Sorry just noticed this now. It was an optimization and never changed the resulting execution. I should have put a comment.

PR #70, opened 2019-10-09
US110195 Add assignment instructions methods

ryantmer:
These allow fetching and editing of the `instructions` on an assignment entity. There are three different representations of the instructions, all of which are made available:

- `instructionsPlaintext`: self-explanatory, this is the HTML-stripped version of the instructions (not really that useful)
- `instructionsHtml`: HTML version of the instructions for displaying the assignment instructions
- `instructionsEditorHtml`: HTML version of the instructions for usage in editors

An example of the difference between the latter two is that an `{orgUnitId}` string template in the former will show as e.g. "6606", whereas in the latter it will just be "{orgUnitId}".

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -1,6 +1,7 @@
 'use strict';
 
 import { Entity } from '../../es6/Entity.js';
+import { Actions, Rels } from '../../hypermedia-constants';
 
 /**
  * AssignmentEntity class representation of a d2l Assignment.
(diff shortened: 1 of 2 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -9,12 +10,52 @@ export class AssignmentEntity extends Entity {
...
+	canEditInstructions() {
+		const instructionsEntity = this._getInstructionsEntity();
+		return instructionsEntity
+			&& instructionsEntity.hasActionByName(Actions.assignments.updateInstructions);
+	}
+
+	getSaveInstructionsAction() {
+		const instructionsEntity = this._getInstructionsEntity();
+		return instructionsEntity
+			&& instructionsEntity.getActionByName(Actions.assignments.updateInstructions);
```
mdulat: Should we be checking `hasActionByName` before doing `getActionByName`?

ryantmer: Not necessarily, and I'd err on the side of not doing it - for the most part it's the same, but the result of `getSaveInstructionsAction` I think is more-normal this way for actions that don't exist.

If we add the `hasActionByName` check and the action doesn't exist,

```js
const saveAction = getSaveInstructionsAction() // == false
```

whereas the current approach results in

```js
const saveAction = getSaveInstructionsAction() // == undefined
```

which seems more predictable to me. An `undefined` action is pretty clearly "there is no such action", whereas a `false` action is less-clear to me (albeit I'll admit, still fairly obvious). That said, I'm open to be convinced otherwise, as I realize this is a minor thing (but one that has come up before, hence having this position/explanation ready to go 😄)

mdulat: Sure, makes sense.

PR #72, opened 2019-10-15
basic activity usage entity

jlai-d2l:
- basic implementation of `ActivityUsage` with the understanding that this will be expanded to support the specialized implementations of activities.
- this contains some of the common links in `ActivityUsage` and unblocks Learning Paths


src/activities/ActivityUsageEntity.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
```
dbatiste: I don't _think_ this should be need here, since "strict" mode is default for ES6 modules.

PR #73, opened 2019-10-15
US110195 Modify getRichTextEditorConfig

ryantmer:
Merged the previous version of this a little too quickly - the richtext editor config should be a sub-entity of the richtext instructions entity, not on the assignment itself.


ryantmer:
Other difficulties came up with this - if we put the richtext config on the instructions sub-entity, this becomes a problem when there aren't any instructions, as there is no instructions sub-entity...

ryantmer:
Solution: always include the instructions sub-entity - updating the LMS to do this (and just have blank instructions when applicable), which means this change is valid once more.

PR #74, opened 2019-10-17
US111258 - Add ability to know how many elements are in the ConsortiumTokenCollectionEntity

svanherk:
This is the second of the three PRs, related to https://github.com/BrightspaceUI/navigation/pull/89.  Since the org tabs are all loaded and added asynchronously, I need a way to know how many tabs we can be expecting, or at least how many tokens we are going to try.  Once we have that, I can check if each of those are done loading, and _then_ tell the scrollbar to start moving.

@ctwomey1 Is this the best way to do this?  Is there an edge scenario I'm missing?

### DIFF
--- a/src/consortium/ConsortiumTokenCollectionEntity.js
+++ b/src/consortium/ConsortiumTokenCollectionEntity.js
@@ -21,6 +21,14 @@ export class ConsortiumTokenCollectionEntity extends Entity {
 		}
 	}
 	/**
+	 *
+	 * @returns length of array of siren consortium entities
+	 */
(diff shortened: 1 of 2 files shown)


ctwomey1:
It seems reasonable.  I added my comments to https://github.com/BrightspaceHypermediaComponents/organizations/pull/129.  I think there's going to be some jankyness around how we are implementing the cache as well as the fact that errors will potentially throw the meaning of this number off.

Let's have a quick zoom call tomorrow to just talk about it real quick.

svanherk:
Closing this for now, since it looks like we can get away without it.

PR #77, opened 2019-10-22
adding edit link to activity usage

jlai-d2l:
[US111291](https://rally1.rallydev.com/#/39602890252d/detail/userstory/342352818840)

### DIFF
--- a/src/hypermedia-constants.js
+++ b/src/hypermedia-constants.js
@@ -162,6 +162,9 @@ export const Rels = {
 	},
 	Users: {
 		canonicalUser: 'https://users.api.brightspace.com/rels/canonical-user'
+	},
+	IANA: {
+		edit: 'edit'
 	}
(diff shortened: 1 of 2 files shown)


src/hypermedia-constants.js
```diff
@@ -162,6 +162,9 @@ export const Rels = {
 	},
 	Users: {
 		canonicalUser: 'https://users.api.brightspace.com/rels/canonical-user'
+	},
+	IANA: {
+		edit: 'edit'
```
AllanKerr: I believe we want to be moving towards having `rels` and `classes` defined directly in the entity that uses them rather than in `hypermedia-constants`. That being said, since IANA Rels are a set of standardized relationships I think it makes sense to keep them here. @m6jones or @ctwomey1 _may_ have some additional context.

PR #78, opened 2019-10-24
Add SirenAction ES6 module

ryantmer:
This is basically a copy of the existing `siren-fetch-mixin-lit`, with the intent of replacing it entirely. (It will be removed entirely in a subsequent PR.) The intent of this ES6 module is to be able to do "Siren things" anywhere - right now limited to just `performSirenAction`, but potentially with more functionality to come. This can then be dropped into any class that extends `Entity`.

One difference between this and `siren-fetch-mixin-lit` is that `SirenAction.performSirenAction` doesn't emit `d2l-siren-entity-save-*` events. An HTMLElement is required to fire these events, and passing it in was weird and blurred the line of what exactly `siren-sdk` does - so instead, consumers of `SirenAction.performSirenAction` will be responsible for their own start/end/error behaviour, which is easily implemented. This keeps a clearer line between "Siren things" (which `siren-sdk` does) and ancillary behaviour (which is the responsibility of the consumer).

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -3,6 +3,7 @@ import { Actions, Classes, Rels } from '../hypermedia-constants';
 import { OrganizationEntity } from '../organizations/OrganizationEntity.js';
 import { UserActivityUsageEntity } from '../enrollments/UserActivityUsageEntity.js';
 import { ActivityUsageCollectionEntity } from './ActivityUsageCollectionEntity.js';
+import { performSirenAction } from '../es6/SirenAction.js';
 
 /**
  * ActivityUsageEntity class representation of a d2l activity usage.
@@ -73,5 +74,20 @@ export class ActivityUsageEntity extends Entity {
 	saveDueDateAction() {
 		const dueDate = this._getDueDateSubEntity();
 		return dueDate && dueDate.getActionByName(Actions.activities.update);
+	}
+
+	async setDueDate(dateString) {
+		const dueDate = this.canEditDueDate && this._getDueDateSubEntity();
+		if (!dueDate) {
+			return;
+		}
+
+		const action = dueDate.getActionByName(Actions.activities.update);
+		if (!action) {
+			return;
+		}
+
+		const fields = [{ name: 'dueDate', value: dateString }];
+		await performSirenAction(this._token, action, fields);
 	}
 }
--- a/src/es6/SirenAction.js
+++ b/src/es6/SirenAction.js
@@ -1,0 +1,161 @@
+import 'd2l-fetch/d2l-fetch.js';
+import 'd2l-polymer-siren-behaviors/store/entity-store.js';
+import 'd2l-polymer-siren-behaviors/store/action-queue.js';
+import SirenParse from 'siren-parser';
+
+const _getSirenFields = function(action) {
+	const url = new URL(action.href, window.location.origin);
+	const fields = [];
+	if (action.method === 'GET' || action.method === 'HEAD') {
+		for (const param in url.searchParams.entries()) {
+			fields.push({ name: param[0], value: param[1] });
+		}
+	}
+
+	if (action.fields && action.fields.forEach) {
+		action.fields.forEach(function(field) {
+			if (field.value === undefined) {
+				return;
+			}
+			// if the field is specified multiple times, assume it is intentional
+			fields.push({ name: field.name, value: field.value });
+		});
+	}
+	return fields;
+};
+
+const _getEntityUrl = function(action, fields) {
+	if (!action) {
+		return null;
+	}
+
+	let url = new URL(action.href, window.location.origin);
+
+	fields = fields || _getSirenFields(action);
+	if (action.method === 'GET' || action.method === 'HEAD') {
+		const params = _createURLSearchParams(fields);
+		url = new URL(url.pathname + '?' + params.toString(), url.origin);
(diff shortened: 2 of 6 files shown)


src/es6/SirenUtil.js
```diff
@@ -0,0 +1,198 @@
...
+const _getSirenFields = function(action) {
+	const url = new URL(action.href, window.location.origin);
+	const fields = [];
+	if (action.method === 'GET' || action.method === 'HEAD') {
+		for (const param in url.searchParams.entries()) {
+			fields.push({ name: param[0], value: param[1] });
+		}
+		// Disable URLSearchParams until they are fully supported (i.e. Edge)
+		/*
+		} else if (window.URLSearchParams && action.type === 'application/x-www-form-urlencoded') {
```
ctwomey1: prob should just delete this code or do capability testing if we wanted to use it and gracefully fallback

src/es6/SirenUtil.js
```diff
@@ -0,0 +1,198 @@
...
+	if (sendSaveEvent) {
+		self.dispatchEvent(new CustomEvent('d2l-siren-entity-save-start', {
+			bubbles: true,
+			composed: true
+		}));
+	}
+	return window.d2lfetch.fetch(href, opts)
+		.then(function(resp) {
+			if (sendSaveEvent && resp.ok) {
+				self.dispatchEvent(new CustomEvent('d2l-siren-entity-save-end', {
```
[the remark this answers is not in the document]
ctwomey1: Alternatively, I wonder if you really care about the element that initiated the event or that the entity id has saved it.

Like do i care that course tile x is done saving? or do i care that http://path.to.course.tile.thing is done saving?

src/es6/SirenUtil.js
```diff
@@ -0,0 +1,198 @@
...
+			const entity = SirenParse(result.body);
+			return Promise.all(linkRequests).then(function() {
+				return window.D2L.Siren.EntityStore.update(url.href, token, entity);
+			});
+		});
+};
+
+const performSirenAction = function(action, fields, immediate) {
+	const self = this; // eslint-disable-line no-invalid-this
+	return window.D2L.Siren.EntityStore.getToken(self.token)
```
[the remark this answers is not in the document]
ctwomey1: I do like passing the token in.  Might feel weird to always do it.  If we migrated it to a class, we could be building this into a constructor instead?  then the this we are referencing is the right this?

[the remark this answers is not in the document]
ryantmer: Nah, I think it's good since it makes this stuff even more stateless!

src/es6/SirenUtil.js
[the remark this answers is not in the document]
ryantmer: In our current usage, it's the element as that's the way the listeners are set up - but that could be changed. In reality, we don't actually care where these events are coming from anyway, since there's just the one global-esque autosave component on the page listening for these events. That said, `dispatchEvent` is from `HTMLElement`, so either this utility class needs to extend `LitElement`, or I'll need to figure out another way 🤔

src/es6/SirenUtil.js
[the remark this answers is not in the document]
ryantmer: Er, when you "migrated it to a class" - you mean make `SirenUtils` a class instead, that we instantiate with a token and... maybe a backing `HTMLElement`?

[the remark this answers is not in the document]
ctwomey1: After talking offline  @ryantmer gonna do a cool thing with callbacks and stuff

src/es6/SirenAction.js
```diff
@@ -0,0 +1,161 @@
+import 'd2l-fetch/d2l-fetch.js';
```
ryantmer: Hmmm - this is now basically a copy of https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/siren-action-behavior.js - should they/can they be reconciled somehow? Maybe I should just call into that?

ctwomey1: probably yes

ryantmer: I think I'll not do this for now - the Polymer behaviour generates the now-removed events that `siren-fetch-mixin-let` was also dispatching, which would mean a refactor of that library as well, which I'm not super jazzed about doing. It would also introduce another ES6 -> Polymer dependency, which I'd rather avoid.

In other words, "yes it's a good idea, but not now".

ctwomey1: Sounds goodish.  Are you going to remove the siren-fetch-mixin-lit soonish or no?

ryantmer: Yeah, I was going to do a follow-up PR for it, so that I can do this additive change, then change the consumer over, then do the (technically-breaking) removal of the old stuff. I'm going to be naughty and not bump the major version though, since it's very-new code and is only being used by us (and doing the major version bump everywhere would be bleh).

src/activities/ActivityUsageEntity.js
```diff
@@ -74,4 +75,19 @@ export class ActivityUsageEntity extends Entity {
...
+			return;
+		}
+
+		const action = dueDate.getActionByName(Actions.activities.update);
+		if (!action) {
+			return;
+		}
+
+		const fields = [{ name: 'dueDate', value: dateString }];
+		return performSirenAction(this._token, action, fields);
```
ctwomey1: This might be weird to return a raw siren response?

PR #79, opened 2019-10-25
Initial commit for OrganizationAvailabilitySet and OrganizationAvailability

wongvincent:
To be used by organizations: https://github.com/BrightspaceHypermediaComponents/organizations/pull/132

### DIFF
--- a/src/organizations/OrganizationAvailabilitySetEntity.js
+++ b/src/organizations/OrganizationAvailabilitySetEntity.js
@@ -1,0 +1,48 @@
+'use strict';
+
+import { Entity } from '../es6/Entity.js';
+import { Rels, Classes, Actions } from '../hypermedia-constants';
+import { OrganizationEntity } from './OrganizationEntity.js';
+
+/**
+ * OrganizationAvailabilitySetEntity class representation of a d2l OrgUnitAvailabilitySet.
+ */
+export class OrganizationAvailabilitySetEntity extends Entity {
+	_getCurrentOrgUnitAvailabilityHref() {
+		const currentOrganizationHref = this.getOrganizationHref();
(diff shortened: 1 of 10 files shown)


src/organizations/OrganizationAvailabilitySetEntity.js
```diff
@@ -0,0 +1,48 @@
...
+
+import { Entity } from '../es6/Entity.js';
+import { Rels, Classes, Actions } from '../hypermedia-constants';
+import { OrganizationEntity } from './OrganizationEntity.js';
+
+/**
+ * OrganizationAvailabilitySetEntity class representation of a d2l OrgUnitAvailabilitySet.
+ */
+export class OrganizationAvailabilitySetEntity extends Entity {
+	_getCurrentOrgUnitAvailabilityHref() {
```
wongvincent: this function is kind of gross, but i need a way to determine if there is an explicit entity in the OrganizationAvailabilitySet that matches the currentOrgUnit.

ctwomey1: Please don't parse the orgunitid from the href. Every time we've done this it eventually caused us grief.

Let's talk Monday about alternative ways to do this

mdgbayly: Yeh, would agree we don't want to be parsing URLs to get org unit ids.
Don't we have the href for the current org unit from the intial response?
So can't we pass it into these entity helpers from the UI?

mdgbayly: Oh - you are using that... So isn't the problem that you need to use the `getOrganizationHref` against the set of entities that you load from the cache primer? i.e. compare the org unit href that you get from `getOrganizationHref` here, with the org unit href's in each of the `OrganizationAvailabilityEntity`s?

Is there something about the way the UI is being built that makes that hard?

PR #80, opened 2019-10-28
US110198 Add methods for submission type

ryantmer:
This simply surfaces the current submission type for assignment, the list of submission type options, and a method to update the current submission type. Note that this doesn't include full support for completion type (coming soon), which means that some options (On Paper and Observed in Person) will always use the default completion type of "Automatically on due date".

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -97,4 +97,57 @@ export class AssignmentEntity extends Entity {
 
 		return this._entity.getLinkByRel(Rels.Activities.activityUsage).href;
 	}
+
+	submissionType() {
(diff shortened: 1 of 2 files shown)


mdulat:
I think they initially wanted to use default values for `completionType`, but maybe if the latter is coming soon it doesn't matter?

ryantmer:
Ooh, okay could be. There isn't a default value, but we could just use "the first value" or something. I'll bring it up tomorrow!

ryantmer:
Ended up just using "Automatically on due date" as the default - kind of a moot point since we'll be adding proper completion type support sooner rather than later anyway, so once the `select` list is added for that, it can be hooked up pretty easily.

PR #83, opened 2019-10-29
US111410 Add support for adding due date

ryantmer:
This uses the new `start-add-new` workflow to set an initial due date if one isn't already set. This is exposed via the same `setDueDate` method, meaning the consumer is blind to this change, and the `siren-sdk` will either set or update the due date as required.

This also removes a method that returns a Siren action, as this was added in error - `siren-sdk` should not expose Siren objects.

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -69,8 +69,7 @@ export class ActivityUsageEntity extends Entity {
 	canEditDueDate() {
 		const dueDate = this._getDueDateSubEntity();
 		return (dueDate && dueDate.hasActionByName(Actions.activities.update))
-			|| this._entity.hasActionByName(Actions.activities.startAddNew)
-			|| false;
+			|| this._entity.hasActionByName(Actions.activities.startAddNew);
 	}
 
 	async setDueDate(dateString) {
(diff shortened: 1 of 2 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -68,21 +68,36 @@ export class ActivityUsageEntity extends Entity {
...
 		const dueDate = this._getDueDateSubEntity();
-		return (dueDate && dueDate.hasActionByName(Actions.activities.update)) || false;
-	}
-
-	saveDueDateAction() {
-		const dueDate = this._getDueDateSubEntity();
-		return dueDate && dueDate.getActionByName(Actions.activities.update);
+		return (dueDate && dueDate.hasActionByName(Actions.activities.update))
+			|| this._entity.hasActionByName(Actions.activities.startAddNew)
+			|| false;
```
mdulat: What's the purpose of `|| false`? Wouldn't the statement just be false if the other components are both false?

ryantmer: It's just type coercion due to the name of the method - since it's `can*`, I infer that it should return `true` or `false` specifically, rather than `undefined` (which it would without the `|| false`)

ryantmer: ...except that it calls `hasActionByName` which pretty sure returns a boolean anyway. So nvm, this is not required and I'll remove it!

PR #85, opened 2019-10-31
Add support for editing draft status

kieranderson:
This adds the ability to update the draft status of an activity.


src/activities/ActivityUsageEntity.js
```diff
@@ -104,4 +104,22 @@ export class ActivityUsageEntity extends Entity {
...
+		return this._entity && this._entity.hasActionByName(Actions.activities.updateDraft);
+	}
+
+	async setDraftStatus(isDraft) {
+		if (!this.canEditDraft()) {
+			return;
+		}
+
+		let updateDraftAction = this._entity.getActionByName(Actions.activities.updateDraft);
+		const fields = [{ name: Classes.activities.draft, value: isDraft }];
```
ryantmer: I think this is better to just leave as the string `'draft'`, although I don't have a strong preference - mostly, I see this as "the class happens to be the same as the field name", not that they are necessarily the same.

That said, we also haven't really library-ified field names at all; no real reason for that as far as I know, just hasn't ever been done ☹️

src/activities/ActivityUsageEntity.js
```diff
@@ -104,4 +104,22 @@ export class ActivityUsageEntity extends Entity {
 		const fields = [{ name: 'dueDate', value: dateString }];
 		await performSirenAction(this._token, action, fields);
 	}
+
+	canEditDraft() {
+		return this._entity && this._entity.hasActionByName(Actions.activities.updateDraft);
+	}
+
+	async setDraftStatus(isDraft) {
```
ctwomey1: Some jsdocs would be great here

ryantmer: Ah shit I haven't done that at all - I'll make a note to do so in the stuff I've added

ctwomey1: no worries.  It'd just be good to get the docs running so people could reference what the sdk can do without digging in code

kieranderson: Ran the generation for the ActivityUsageEntity, let me know if there is other steps in the workflow here.

ctwomey1: Adding detail to the jsdoc for public methods would make it easier for consumers to reason  about the function call.  Something similar to https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/consortium/ConsortiumTokenEntity.js 

This is an example of how to use jsdocs in general https://devdocs.io/jsdoc/howto-es2015-classes

PR #87, opened 2019-11-12
Adding support for specialization entity and new mixins for consisten…

jlai-d2l:
…t activity properties.


src/mixin/describable-entity-mixin.js
```diff
@@ -0,0 +1,27 @@
+import { performSirenAction } from '../es6/SirenAction.js';
+
+export const DescribableEntityMixin = superclass => class extends superclass {
+
+	_isDescribableEntity() {
```
AllanKerr: I think it makes sense to make these checks public. The reasoning is that it lets the component thats using it decide whether the mixin behavior is optional or raise an error of some form if its required for the component to function.

src/hypermedia-constants.js
```diff
@@ -36,6 +36,7 @@ export const Rels = {
 	semesters: 'https://api.brightspace.com/rels/semesters',
 	sequence: 'https://api.brightspace.com/rels/sequence',
 	sorts: 'https://api.brightspace.com/rels/sorts',
+	specialization: 'https://api.brightspace.com/rels/specialization',
```
m6jones: I'd prefer if new ones stay in the file where they are used so we can start getting away from this file.

src/mixin/describable-entity-mixin.js
```diff
@@ -0,0 +1,27 @@
+import { performSirenAction } from '../es6/SirenAction.js';
+
+export const DescribableEntityMixin = superclass => class extends superclass {
```
m6jones: I'd like to hear more about what lead to this pattern. 

The plan is to avoid specific mixins but plans sometimes don't work in the real world.

m6jones: Oh, I see how they are being used. Interesting. 

I'd want to separate them from the mixins that are framework specific.

src/entityAddons/named-entity-mixin.js
```diff
@@ -0,0 +1,27 @@
...
+	}
+
+	getName() {
+		if (this.isNamedEntity()) {
+			return super.entity().properties.name;
+		}
+		return;
+	}
+
+	async setName(name) {
```
m6jones: This being async shouldn't be required. But leave it for now.

m6jones: Mainly, wanted to point this out if we look back on this pr as an example

PR #88, opened 2019-11-12
Checklist as an Activity

wongvincent:
**This is not ready for review.**
Some code we can use in the future when the other changes for Simplified Interface Checklist (SIC) are made.


wongvincent:
lol

PR #95, opened 2019-11-27
Rvinnakota/us111803 annotations can be added to an assignment

rohitvinnakota:
LMS PR here:
https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/5925/overview

Rally story here:
https://rally1.rallydev.com/#/110294864140d/detail/userstory/349211942424

PR in AssignmentEditorDetail in BrightspaceHypermediaComponents/activities to follow

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -1,7 +1,7 @@
 'use strict';
 
 import { Entity } from '../../es6/Entity';
-import { Actions, Rels } from '../../hypermedia-constants';
+import { Actions, Rels, Classes } from '../../hypermedia-constants';
 import { performSirenAction } from '../../es6/SirenAction';
 
 /**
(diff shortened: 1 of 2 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -258,4 +258,28 @@ export class AssignmentEntity extends Entity {
 		];
 		await performSirenAction(this._token, action, fields);
 	}
+
+	annotationToolsAvailable() {
```
ryantmer: I'd recommend `getAvailableAnnotationTools` here - just to keep consistent the `get` nomenclature elsewhere

src/activities/assignments/AssignmentEntity.js
```diff
@@ -258,4 +258,28 @@ export class AssignmentEntity extends Entity {
...
+		return this._entity.getSubEntityByClasses(['annotations', 'enabled']);
```
ryantmer: I'm not 100% sure, but I _suspect_ you actually want `getSubEntitiesByClasses` (plural) here? `getSubEntityByClasses` (singular) will just return the first entity that matches the requirements, whereas `getSubEntitiesByClasses` (plural) will return an array of all entities that match the requirements.

src/activities/assignments/AssignmentEntity.js
```diff
@@ -258,4 +258,28 @@ export class AssignmentEntity extends Entity {
...
+	}
+
+	canSeeAnnotations() {
```
ryantmer: Adding a quick JSDoc to each of these would be appreciated! We're not perfect on it, but all new things should have them.

src/activities/assignments/AssignmentEntity.js
ryantmer: Applies elsewhere too, but there is a `Classes` in `hypermedia-constants` that these could be added to.

rohitvinnakota: I just want the single `annotations` sub-entity that contains the action and the current state of annotations for the assignment. There should be only one sub-entity with an `annotations` class per assignment.

src/activities/assignments/AssignmentEntity.js
rohitvinnakota: Is `_getAvailableAnnotationTools` preferred over `getAvailableAnnotationTools`?

ryantmer: Depends - is this being used by your consumer as well? (The underscore is just being used to denote private methods by convention, really.)

rohitvinnakota: Yes it is to set the checkbox state, so `getAvailableAnnotationTools`?

ryantmer: Sounds right to me!

src/activities/assignments/AssignmentEntity.js
ppaskaris-d2l: Shouldn't we ideally look for things by relationship and not their class?

ryantmer: Actually, that's a better point. `getSubEntityByClass` can generally be replaced with a call to `getSubEntityByRel`.

PR #97, opened 2019-11-28
Adding editing of start/end date to ActivityUsageEntity

kieranderson:


docs/activities_ActivityUsageEntity.js.html
```diff
@@ -36,6 +36,8 @@ <h1 class="page-title">Source: activities/ActivityUsageEntity.js</h1>
 /**
  * ActivityUsageEntity class representation of a d2l activity usage.
  */
```
ppaskaris-d2l: this file is generated noise, yeah?

kieranderson: It's generated yeah, from the comments for the JSdocs

src/activities/ActivityUsageEntity.js
```diff
@@ -71,31 +71,105 @@ export class ActivityUsageEntity extends Entity {
...
+	async _setDate(dateValue, dateClass, dateField) {
+		if (!this._canEditDate(dateClass)) {
 			return;
 		}
 
 		let action;
-		const dueDateEntity = this._getDueDateSubEntity();
-		if (dueDateEntity) {
-			action = dueDateEntity.getActionByName(Actions.activities.update);
+		const dateEntity = this._getDateSubEntity(dateClass);
```
ppaskaris-d2l: Still irks me to see fetching sub entities by class instead of rel. Surely the class ought to be "date" and the rel ought to be "due" 🤷‍♀

ppaskaris-d2l: Oh, the actions are different anyway, so one kind of date doesn't quack like a different kind of date 🤔

src/activities/ActivityUsageEntity.js
```diff
@@ -86,41 +86,105 @@ export class ActivityUsageEntity extends Entity {
...
+		await this._setDate( dateValue, Classes.dates.endDate, 'endDate');
+	}
+
+	_getDateSubEntity(dateClass) {
+		return this._entity
+			&& this._entity.getSubEntityByClass(dateClass);
+	}
+
+	/**
+	 * @param {string} dateClass The class for the date entity
```
ryantmer: Not a biggie, but JSDoc-ing these private-ish methods will surface them in the docs and potentially make people use them - I'd say either make them intentionally public-ish, or just leave them undocumented. (Admittedly though, this is a feeble attempt to add access modifiers to JS, which... nope.)

src/activities/ActivityUsageEntity.js
ryantmer: Yeah, the field name for each date is the name of the date - you are correct that it probably could have just been `date`.

src/activities/ActivityUsageEntity.js
```diff
@@ -86,41 +86,95 @@ export class ActivityUsageEntity extends Entity {
...
+		return this._canEditDate(Classes.dates.dueDate);
 	}
 
 	/**
  	 * Updates the due date of the activity usage entity to the date specified
-	 * @param {string} dateString Date string to set as the due date, or empty string to clear the due date
+	 * @param {string} dateValue Date string to set as the due date, or empty string to clear the due date
+	 */
+	async setDueDate(dateValue) {
+		await this._setDate(dateValue, Classes.dates.startDate, 'dueDate');
```
mdulat: Should the 2nd param be `Classes.dates.dueDate`?

ryantmer: o dang, so it should - your eyes are better than mine!

kieranderson: It sure should be!

src/activities/ActivityUsageEntity.js
mdulat: well... the endpoint for updating each date is the same (the usage endpoint), so the field names have to be unique for each date.

ryantmer: 🤦‍♂ Right, totally correct. For some reason I thought we could have them on the same endpoint, and also the same field name, and... idk, magically know which one to update? Not sure why I thought that 😛

PR #98, opened 2019-11-28
Add assignment release conditions

ppaskaris-d2l:
[US109698](https://rally1.rallydev.com/#/detail/userstory/332018306084?fdp=true)

This PR adds the assignment release conditions stuff.

### DIFF
--- a/src/hypermedia-constants.js
+++ b/src/hypermedia-constants.js
@@ -68,6 +68,7 @@ export const Rels = {
 	// Assignments
 	Assignments: {
 		instructions: 'https://assignments.api.brightspace.com/rels/instructions',
+		releaseConditions: 'https://assignments.api.brightspace.com/rels/release-conditions',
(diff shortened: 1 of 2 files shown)


src/hypermedia-constants.js
```diff
@@ -68,6 +68,7 @@ export const Rels = {
 	// Assignments
 	Assignments: {
 		instructions: 'https://assignments.api.brightspace.com/rels/instructions',
+		releaseConditions: 'https://assignments.api.brightspace.com/rels/release-conditions',
```
ppaskaris-d2l: I'm thinking that since this is a preliminary crappy version of it, I'm going to pick a different rel so as to not steal the good one.

ppaskaris-d2l:
Apologies all for the spam, Kieran has pointed out that I've made the error of putting it on the "assignment entity" instead of the "activity usage entity". I will redo it and re-open later.

PR #104, opened 2019-12-03
US111919 Add assignments turnitin dialog opener

ppaskaris-d2l:
[US111919](https://rally1.rallydev.com/#/detail/userstory/350454972296?fdp=true)

This PR adds the assignment turnitin dialog stuff. Also, apparently `this._entity` can be weird, so I added a check for that back into the release conditions dialog stuff too.

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -60,6 +60,10 @@ export class ActivityUsageEntity extends Entity {
 	 */
 	canEditReleaseConditions() {
 
+		if (!this._entity) {
+			return false;
+		}
+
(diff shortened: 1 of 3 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -60,6 +60,10 @@ export class ActivityUsageEntity extends Entity {
 	 */
 	canEditReleaseConditions() {
 
+		if (!this._entity) {
```
ctwomey1: You can condense this like https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/f12c4f907578baea38e24e61d9b93e22b64820af/src/organizations/OrganizationEntity.js#L28

ideally we'd be using https://github.com/tc39/proposal-optional-chaining to make that check not a thing, but you know how it is...

PR #106, opened 2019-12-06
subEntity load promise

m6jones:
Did you ever have the problem where you wanted to use lit's awesome async states but just didn't have a promise when all the entities have loaded?

Well, now you can use `subEntitiesLoaded()` which is a promise that will load when all the sub entities have loaded. It even deals with multiple levels too!

For example, consider having a component root hypermedia entity be an ActivityUsage. That activity usage turns out to have a collection. This is how we would let litElement know when the list of items has changed after everything has loaded:
```
_onActivityUsageChange(usage) {
	usage.onSpecializationChange(NamedEntityMixin(DescribableEntityMixin(SimpleEntity)), (specialization) => {
		this._specialization = specialization;
		this._name = specialization.getName();
		this._description = specialization.getDescription();
	});

	usage.onActivityCollectionChange(async collection => {
		collection.onItemsChange((item, index) => {
			item.onActivityUsageChange((usage) => {
				usage.onOrganizationChange((organization) => {
					this._items[index] = organization;
					if (this._updateSingles) {
						this.requestUpdate('_items', []);
					}
				});
			});
		});

		await collection.subEntitiesLoaded();
		this.requestUpdate('_items', []);
		this._updateSingles = true;
	});
}
```

### DIFF
--- a/src/es6/Entity.js
+++ b/src/es6/Entity.js
@@ -95,9 +95,15 @@ export class Entity {
 				this._subEntities.set(source, entity);
 				onChange(entity, error);
 				if (error) {
-					reject();
+					reject && reject();
+					resolve = null;
+					reject = null;
 				} else {
-					Promise.all(entity._subEntitiesLoadStatus).then(resolve);
+					Promise.all(entity._subEntitiesLoadStatus).then(() => {
+						resolve && resolve();
+						resolve = null;
+						reject = null;
+					});
 				}
 			});
 		}));
@@ -126,7 +132,10 @@ export class Entity {
 			entityFactory(entityType, href, this._token, (entity) => {
 				this._subEntities.set(href, entity);
 				onChange(entity);
-				Promise.all(entity._subEntitiesLoadStatus).then(resolve);
+				Promise.all(entity._subEntitiesLoadStatus).then(() => {
+					resolve && resolve();
+					resolve = null;
+				});
 			}, entity);
 		}));
 	}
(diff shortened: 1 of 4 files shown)


src/es6/Entity.js
```diff
@@ -37,6 +38,9 @@ export class Entity {
 
 		return this._entity.getLinkByRel('self').href;
 	}
+	async subEntitiesLoaded() {
```
dkong-d2l: This could just be a function that returns a promise...

src/es6/Entity.js
```diff
@@ -82,10 +89,24 @@ export class Entity {
...
-		entityFactory(entityType, source, this._token, (entity) => {
-			this._subEntities.set(source, entity);
-			onChange(entity);
-		});
+
+		this._subEntitiesLoadStatus.push(new Promise((resolve, reject) => {
+			entityFactory(entityType, source, this._token, (entity, error) => {
+				this._subEntities.set(source, entity);
+				onChange(entity, error);
+				if (error) {
```
dkong-d2l: This code block could be refactored...

src/es6/Entity.js
```diff
@@ -82,10 +89,24 @@ export class Entity {
...
+					reject && reject();
+					resolve = null;
+					reject = null;
+				} else {
+					Promise.all(entity._subEntitiesLoadStatus).then(() => {
+						resolve && resolve();
+						resolve = null;
+						reject = null;
```
m6jones: Setting these to null because it has been rumoured to cause problems if you do not. I am assuming polyfills get angry. So better be safe.

src/es6/Entity.js
m6jones: Yup!

PR #107, opened 2019-12-11
Add OrganizationCollectionEntity

AllanKerr:
# Changes
1. This pull request adds an `OrganizationCollectionEntity` to represent the response returned from the `OrganizationsController` with support for sorting and paging.
2. Fixed a bug in `SirenAction` where hidden fields are dropped for actions using `GET`.

### DIFF
--- a/src/es6/SirenAction.js
+++ b/src/es6/SirenAction.js
@@ -31,7 +31,11 @@ const _getEntityUrl = function(action, fields) {
 
 	let url = new URL(action.href, window.location.origin);
 
-	fields = fields || _getSirenFields(action);
+	if (fields) {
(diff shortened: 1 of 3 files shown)


src/es6/SirenAction.js
```diff
@@ -31,7 +31,11 @@ const _getEntityUrl = function(action, fields) {
 
 	let url = new URL(action.href, window.location.origin);
 
-	fields = fields || _getSirenFields(action);
+	if (fields) {
```
m6jones: Nit: 
```
fields = fields ? fields = _appendHiddenFields(action, fields) : _getSirenFields(action);
```

PR #108, opened 2019-12-12
Remove Item from collection

m6jones:
Allows us to remove an item from the collection

### DIFF
--- a/src/activities/ActivityUsageCollectionEntity.js
+++ b/src/activities/ActivityUsageCollectionEntity.js
@@ -36,9 +36,7 @@ export class ActivityUsageCollectionEntity extends Entity {
 		this._entity.entities.forEach((entity) => {
 			if (href === entity.getLinkByRel('self').href) {
 				const action = entity.getActionByName(actions.removeActivity);
-				performSirenAction(this._token, action).then((response) => {
-					this.update(response);
-				});
+				performSirenAction(this._token, action);
 			} else {
 				newEntityList.push(entity);
 			}
--- a/src/activities/ActivityUsageCollectionEntity.js
+++ b/src/activities/ActivityUsageCollectionEntity.js
@@ -3,6 +3,11 @@
 import { Entity } from '../es6/Entity.js';
 import { Rels } from '../hypermedia-constants';
 import { ActivityUsageEntity } from './ActivityUsageEntity.js';
+import { performSirenAction } from '../es6/SirenAction.js';
+
+const actions = {
+	removeActivity: 'remove-activity'
+};
 
 /**
  * ActivityUsageCollectionEntity class representation of a d2l activity usage collection.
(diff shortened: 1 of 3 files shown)


src/activities/ActivityUsageCollectionEntity.js
```diff
@@ -1,42 +1,63 @@
...
-	}
-
-	onActivityUsageChange(onChange) {
-		const activityUsageHref = this.activityUsageHref();
-		activityUsageHref && this._subEntity(ActivityUsageEntity, activityUsageHref, onChange);
-	}
-}
+'use strict';
+
+import { Entity } from '../es6/Entity.js';
```
m6jones: The line spacing was wrong :O

src/activities/ActivityUsageCollectionEntity.js
```diff
@@ -1,42 +1,63 @@
...
+		});
+	}
+
+	removeItem(href) {
+		const newEntityList = [];
+		this._entity.entities.forEach((entity) => {
+			if (href === entity.getLinkByRel('self').href) {
+				const action = entity.getActionByName(actions.removeActivity);
+				performSirenAction(this._token, action).then((response) => {
+					this.update(response);
```
m6jones: This is to double check if the action was successful and update to element to make sure it matches the saved version

src/activities/ActivityUsageCollectionEntity.js
```diff
@@ -1,42 +1,63 @@
...
+				});
+			} else {
+				newEntityList.push(entity);
+			}
+		});
+		this._entity.entities = newEntityList;
+		this.update(this._entity);
```
m6jones: This is likely the outcome so lets just render it for a better use experience. When it fails it might suck flipping back but that will not happen much so why not do it quick!

src/activities/ActivityUsageCollectionEntity.js
AllanKerr: Is this necessary? Shouldn't the response automatically trigger the `onChange` callback since it has the same `href`? Then we could call the action directly on the sub-entity rather than having to iterate through comparing `href`s.

m6jones: I think I should check if performSirenAction does something like this already. I sort of assumed it didn't.

m6jones: Oh, so I am doing this iteration so I can delete it from the entity at the same time. So the entity gets updated right away. Instead of waiting for a reply from the server since this action is going to pass 99% of the time we might as well make the experience super quick for those users.

AllanKerr: I could be mistaken but won't the promise not resolve until the new collection entity has been returned so calling update here won't save us any time, it'll just cause a double re-render?

m6jones: It won't re-render if nothing has changed. So we don't need to worry about that.

It does sent two out:
--> happens only on client side so everything is returned super fast and we don't need to wait.
--> Lets send out the action and wait for the response and call render if anything has changed.

m6jones: In the future it be nice to be able to popup an alert saying failed to delete and refresh the page back instead of just refreshing the page back.

m6jones: Off line talk. I miss understood and forgot about the first comment.

PR #109, opened 2019-12-14
Stop d2l-pending-state event propagation

ryantmer:
By allowing the `d2l-pending-state` event to continue bubbling up the DOM, through every `PendingContainerMixin` element, we end up in a situation where the topmost Container can make overarching decisions that it shouldn't necessarily make.

In FACE, this shows when adding a new attachment - the list of attachments emits the `d2l-pending-state` event, and we want to dynamically reload/re-render just the attachment list. However, the event continues to bubble up to the top level, which interprets as "the page is still loading", and switches to show the "Loading..." text, effectively re-rendering the entire page. By stopping the event's propagation, we can have better control over how much of the page re-renders - effectively saying "if any of my children emit `d2l-pending-state`, I can handle the loading state for them".

### DIFF
--- a/src/mixin/pending-container-mixin.js
+++ b/src/mixin/pending-container-mixin.js
@@ -32,6 +32,9 @@ export const PendingContainerMixin = superclass => class extends superclass {
 		}
 		this._hasPendingChildren = true;
 		this.__pendingCount++;
+
+		e.stopPropagation();
+
 		try {
 			await promise;
 		} catch (e) {


dbatiste:
What are your thoughts on switching to use [AsyncContainerMixin](https://github.com/BrightspaceUI/core/tree/master/mixins/async-container)? It is designed to not return to the initial/pending states after reaching its final state unless its reset method is called.  For instance, list loads, then load-more is called... we don't want to return the container view to the loading state. It does however assume that if there are multiple async operations pending, that they overlap (i.e. once there are no pending promises, it changes to its final state).

It also supports a delay - in cases where the pending state duration may be very short, a delay can be used to minimize flashing of the pending state.

One notable difference, `AsyncContainerMixin` listens for `pending-state` event (no `d2l-` prefix).  This was done simply to align with the pending state protocol, in which the vision is that the event name may eventually become more commonly used.  We could align ourselves either way.

ryantmer:
Definitely interesting! In this change, we're bound to the `d2l-` part of the event as that's what the entity store is generating. However, the entity store could emit both events I suppose, for the transition period towards the more standards-y `pending-state` event, and then eventually we could stop using `d2l-pending-state` altogether. It does sound like the don't-return-to-loading-state thing is basically the same as what I'm doing here and in [the related PR](https://github.com/BrightspaceHypermediaComponents/activities/pull/547).

dbatiste:
I think I was on the fence about the event name.  I think originally I was advocating for the `d2l-pending-state` event, but then when I read Justin's comments about the idea of different component providers using it, I thought I'd just stick with `pending-state`.  That being said, I think we'd be safe to stick with `d2l-pending-state`... just need to update the async container mixin in core, and the async directive.  Which leads to... maybe we should have a reusable helper method for dispatching this event. It could perhaps be included with the async directive file to avoid an additional HTTP request.

ryantmer:
Fair! Yeah the existing PendingContainer with `d2l-pending-state` is used in a few places, so I guess just comes down to which is easier.

Unrelated to that, this change actually isn't needed anyway, so closing it.

mdgbayly:
> What are your thoughts on switching to use [AsyncContainerMixin]

Yep, was not aware that existed until now, but it seems like it would be reasonable for us to switch to using it at some point.

dbatiste:
Cool.  Yeah it's a relatively new addition in core.  It was initially created so that the new `d2l-dialog` containing async content could initially size itself (and provide a default spinner) once the content is loaded, and defer placing focus.

m6jones:
I am looking forward to using the async container. It works really well with this: https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/es6/Entity.js#L41

PR #112, opened 2019-12-18
US112361 Add function to attach temp file

ryantmer:
Since we're using the file selector dialog in FACE, it takes care of the actual uploading of files. Once they're uploaded by the dialog, we need to attach the existing temp file to the attachment collection, which can be accomplished via the `add-file` action with a `tempFileId` in the POST body.

This is a followup to #111 - probably should have been included there, but didn't yet know what this would look like.

### DIFF
--- a/src/activities/AttachmentCollectionEntity.js
+++ b/src/activities/AttachmentCollectionEntity.js
@@ -108,4 +108,20 @@ export class AttachmentCollectionEntity extends Entity {
 	canAddFileAttachment() {
 		return this._entity.hasActionByName('add-file');
 	}
+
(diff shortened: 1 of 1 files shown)


dbatiste:
What happens downstream from this - i.e. during/after the POST?  Does the temp file get copied to the domain and the reference to the file updated?  Is that logic that already exists?

ryantmer:
It's logic that already existed - as far as I can deduce, the temp file is copied to the domain and a reference to that new file is added to the attachment collection, so when the temp file is deleted, the attachment is still valid (and the file still exists).

PR #115, opened 2019-12-24
Us111489 add to grades

mdulat:


src/activities/ActivityUsageEntity.js
```diff
@@ -264,4 +264,126 @@ export class ActivityUsageEntity extends Entity {
...
+	 * @returns {string} Score out of value of the activity usage
+	 */
+	scoreOutOf() {
+		if (this.canEditScoreOutOf()) {
+			const action = this._getScoreOutOfAction();
+			if (action.hasFieldByName('scoreOutOf')) {
+				return action.getFieldByName('scoreOutOf').value;
+			}
+		}
+		return this._getScoreOutOfEntity().properties.scoreOutOf;
```
ryantmer: Kind of an edge case concern, but `_getScoreOutOfEntity` can return `null` (if `getSubEntity` doesn't return anything), which would cause `.properties` to blow up.

src/activities/ActivityUsageEntity.js
```diff
@@ -264,4 +264,126 @@ export class ActivityUsageEntity extends Entity {
...
+		if (this.canEditScoreOutOf()) {
+			const action = this._getScoreOutOfAction();
+			if (action.hasFieldByName('gradeType')) {
+				const gradeTypes = action.getFieldByName('gradeType').value.filter(x => x.selected);
+				if (gradeTypes.length > 0) {
+					return gradeTypes[0].title;
+				}
+			}
+		}
+		return this._getScoreOutOfEntity().properties.gradeType || 'Points';
```
ryantmer: The default value here strikes me as odd, but I'm probably just missing context - is this just because "Points" is an always-safe fallback, and because all grades _must_ have a type? (So if `properties.gradeType` is missing for whatever reason, we can 100% assume that's a bug/error in the response and just return "Points"?)

test/activities/ActivityUsageEntity.js
```diff
@@ -0,0 +1,258 @@
...
+		});
+
+		it('can get release conditions url', () => {
+			expect(entity.editReleaseConditionsUrl()).to.equal('/d2l/le/conditionalRelease/6609/dialog/dropboxes/31/openDialog');
+		});
+	});
+
+	describe('Due Date', () => {
+		describe('Can edit', () => {
+			let sandbox, setDueDateSpy;
```
ryantmer: minor thing, the `sandbox` can be defined for the whole file and created/restored in the topmost `beforeEach/afterEach`. Just removes a bit of repeated setup! (albeit not much, admittedly)

src/activities/ActivityUsageEntity.js
```diff
@@ -264,4 +264,126 @@ export class ActivityUsageEntity extends Entity {
 	isPublished() {
 		return this._entity && this._entity.hasClass(Classes.activities.draftPublishedEntity) && this._entity.hasClass(Classes.activities.published);
 	}
+
+	/**
+	 * @returns {string} Score out of value of the activity usage
+	 */
+	scoreOutOf() {
+		if (this.canEditScoreOutOf()) {
```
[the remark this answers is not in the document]
mdulat: This is getting the value from the action if it exists (i.e. if user can edit), and from the properties otherwise. Perhaps this is confusing, I can change it to explicitly check if the action exists.

src/activities/ActivityUsageEntity.js
mdulat: Right, will add a check there...

src/activities/ActivityUsageEntity.js
mdulat: It's for the case when a user first clicks into the points "ungraded" button - at this point the score, grade, grade type all don't exist yet but we want to show this in the UI as the default value that will be used if the user doesn't change it. So now I'm thinking this belongs in the activities repo, not here since it's a UI spec thing.

src/activities/ActivityUsageEntity.js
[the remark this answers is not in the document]
ryantmer: Is there a need to fetch it from the action though? Wouldn't that always be the same as the property anyway, or are there situations where they can differ?

src/activities/ActivityUsageEntity.js
ryantmer: Ah yeah, you mentioned this in your demo I recall. But yeah, I think that would probably make more sense in the UI side :+1:

src/activities/ActivityUsageEntity.js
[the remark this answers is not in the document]
mdulat: Since the action fields are supposed to represent the current state of the form element, it is more "proper" to get the values from there. However if the user has no permissions then there is no action, hence the fallback to properties. This is how we ended up doing it in Activity Feed, to try to keep things as conforming to specs as possible.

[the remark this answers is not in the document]
ryantmer: True, mostly was just thinking you'll either get the same value as the property anyway, or you'll end up falling back to the property value anyway, so didn't really see the need for the extra complexity. But I also don't see there being any harm to doing it one way or the other, so long as its done consistently!

PR #116, opened 2020-01-02
Selfless Entity

m6jones:
Used when there is a sub-entity with no self attribute. Allows the sub-entities of these have a parent that will clean them up.

### DIFF
--- a/src/activities/ActionCollectionEntity.js
+++ b/src/activities/ActionCollectionEntity.js
@@ -1,6 +1,7 @@
 'use strict';
 
 import { Entity } from '../es6/Entity.js';
+import { SelflessEntity } from '../es6/SelflessEntity.js';
 import { Rels } from '../hypermedia-constants';
 import { ActivityUsageEntity } from './ActivityUsageEntity.js';
 import { performSirenAction } from '../es6/SirenAction';
@@ -14,10 +15,13 @@ export class ActionCollectionEntity extends Entity {
 			return;
 		}
 		return this._entity.getSubEntitiesByRel('item').map(item => {
-			const actionItemEntity = new ActionItemEntity();
-			actionItemEntity._entity = item;
+			const actionItemEntity = new ActionItemEntity(this, item);
 			return actionItemEntity;
 		});
+	}
+
+	items() {
+		return this._items();
 	}
 
 	getNextAction() {
@@ -32,7 +36,7 @@ export class ActionCollectionEntity extends Entity {
 
 }
 
-class ActionItemEntity extends Entity {
+class ActionItemEntity extends SelflessEntity {
 	activityUsageHref() {
 		if (!this._entity || !this._entity.hasLinkByRel(Rels.Activities.activityUsage)) {
 			return;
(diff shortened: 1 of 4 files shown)


ctwomey1:
Embedded siren representations are suppose to have a self link

> Root entities and sub-entities that are embedded representations SHOULD contain a links collection with at least one item contain a rel value of self and an href attribute with a value of the entity's URI.
> 
> Sub-entities that are embedded links MUST contain an href attribute with a value of its URI.
from: https://github.com/kevinswiber/siren#entities

I'm not sure we should actually do this

m6jones:
This doesn't mean they all do though :( 

image

I rather have proper clean up on these entities.

m6jones:
We will see about adding self links first.

src/activities/ActionCollectionEntity.js
```diff
@@ -1,25 +1,73 @@
...
 
-import { Entity } from '../es6/Entity.js';
 import { Rels } from '../hypermedia-constants';
 import { ActivityUsageEntity } from './ActivityUsageEntity.js';
 import { performSirenAction } from '../es6/SirenAction';
 
+/**
+ * Abstract Entity class to help create entity classes.
+ */
+class SelflessEntity {
```
ctwomey1: seems to be a dupe of src/es6/selflessentity

src/es6/SelflessEntity.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+/**
+ * Abstract Entity class to help create entity classes.
+ */
+export class SelflessEntity {
+	/**
+	 * Primes the object used by the entityFactory. Should never be called outside.
```
ctwomey1: I thought the intent was that it wasn't used by the entityfactory?

src/es6/SelflessEntity.js
```diff
@@ -0,0 +1,47 @@
...
+	 * @param {*} className Class name you want to see if the entity has attached to it.
+	 */
+	hasClass(className) {
+		return this._entity && this._entity.hasClass(className);
+	}
+	/**
+	 * Get the action with an Action name that is attached to the entity
+	 * @param {*} actionName Action name to get the action attached to the entity.
+	 */
+	getActionByName(actionName) {
```
ctwomey1: https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/es6/Entity.js#L58-L63

The code duplication introduced here feels a bit off.  We should probably consolidate things into something logical?  Could be a simple utility? Worth thinking about

src/es6/SelflessEntity.js
m6jones: Oops! I updated the params but not the initial comment. :(

ryantmer:
Yeah, lots don't - the issue is that a lot aren't addressable, and while having a self link doesn't necessarily mean they are addressable, we don't have an existing pattern to differentiate them. (For example, the due date entity on an activity usage can't be fetched directly.)

m6jones:
Added the new pattern that is cleaned up.
Tasks:
- [ ] Add tests. Note that a test to check if the entity does not go into the store is important

jlai-d2l:
@ctwomey1: Per our conversation, I've updated the documentation to reflect the transient nature of these entities.

PR #118, opened 2020-01-07
US111489 add permission checks for grades

mdulat:


src/activities/ActivityUsageEntity.js
```diff
@@ -373,6 +373,24 @@ export class ActivityUsageEntity extends Entity {
 			&& scoreOutOfEntity.hasActionByName(Actions.activities.scoreOutOf.update);
 	}
 
+	/**
+	 * @returns {bool} Whether or not the inGrades property is present on the score-out-of subentity of the activity usage entity
+	 */
+	canSeeGrades() {
+		const scoreOutOfEntity = this._getScoreOutOfEntity();
+		return scoreOutOfEntity
+			&& typeof scoreOutOfEntity.properties.inGrades !== 'undefined';
```
ryantmer: `scoreOutOfEntity.properties.hasOwnProperty('inGrades')` might be a more intuitive/canonical way to do this (I'm sure there are some weird corner cases where the two differ granted, #BecauseJavascript)

mdulat: I like your way better, updated.

PR #122, opened 2020-01-16
Us111000 grade associations

mdulat:
Adds functionality for associating existing grade items.

### DIFF
--- a/src/activities/GradeCandidateCollectionEntity.js
+++ b/src/activities/GradeCandidateCollectionEntity.js
@@ -1,4 +1,3 @@
-'use strict';
 import { Entity } from '../es6/Entity';
 import { GradeCandidateEntity } from './GradeCandidateEntity';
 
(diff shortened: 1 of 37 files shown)


src/activities/GradeCandidateCollectionEntity.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
ryantmer: Old habits die hard - ES6 modules are all strict mode, so don't need this anymore - but I find myself doing it all the time anyway! 😄

src/activities/GradeCandidateEntity.js
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * Calls the Siren action to associate this grade with the activity
+	 */
+	async associateGrade() {
+		if (!this.canAssociateGrade()) {
+			return;
+		}
+
+		const action = this._entity.getActionByName('associate-grade');
+		await performSirenAction(this._token, action);
```
ryantmer: I take it all the fields here are `hidden` values that are pre-populated yeah?

PR #127, opened 2020-02-03
Kanderson/us109627 edit assignment type

kieranderson:
<a href="https://rally1.rallydev.com/#/detail/userstory/331297385828?fdp=true">US109627</a>

This PR adds the ability to see and set assignment type on assignments to individual or group

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -231,10 +231,10 @@ export class AssignmentEntity extends Entity {
 		if (!action) return;
 
 		
-		const groupTypeId = action.fields[0].value[0].value;
+		const defaultGroupTypeId = action.fields[0].value[0].value;
 		
 		const fields = [
-			{ name: 'groupTypeId', value: groupTypeId },
+			{ name: 'groupTypeId', value: defaultGroupTypeId },
 			{ name: 'folderType', value: 1 }
 		];
 		await performSirenAction(this._token, action, fields);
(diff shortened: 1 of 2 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -144,6 +144,133 @@ export class AssignmentEntity extends Entity {
 		return entity ? entity.properties.url : undefined;
 	}
 
+	/**
+	 * @returns {string} Name of the selected group category for the assignment type
+	 */
+	getAssignmentTypeSelectedGroupCategoryName() {
+		if (!this._entity) return null;
+		const subEntity = this._entity.getSubEntityByRel(Rels.Assignments.folderType);
+		if (!subEntity || !subEntity.properties || !subEntity.properties.groupName) return null;
```
ppaskaris-d2l: All these no-brace ifs seem kind of sneaky, but sneaky can be ok in some cases.

src/activities/assignments/AssignmentEntity.js
```diff
@@ -144,6 +144,133 @@ export class AssignmentEntity extends Entity {
...
+		return subEntity.properties.groupName;
+	}
+
+	/**
+	 * @returns {bool} If the assignment type cannot be changed
+	 */
+	isAssignmentTypeReadOnly() {
```
ppaskaris-d2l: Should this maybe default to `true` given that it is the "you can't do a thing" case?

PR #130, opened 2020-02-07
US113165 Integrate with aggregate save dates action

spandanabaruah:
added new setDates function to call combined dates action. setDates is then used in the overall save function.

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -204,14 +204,14 @@ export class ActivityUsageEntity extends Entity {
 			return;
 		}
 
-		const startDateChanged = typeof startDate !== 'undefined' && startDate !== this.startDate();
-		const dueDateChanged = typeof dueDate !== 'undefined' && dueDate !== this.dueDate();
-		const endDateChanged = typeof endDate !== 'undefined' && endDate !== this.endDate();
+		const startDateChanged = this._hasDateChanged(startDate, this.startDate());
+		const dueDateChanged = this._hasDateChanged(dueDate, this.dueDate());
+		const endDateChanged = this._hasDateChanged(endDate, this.endDate());
 
 		if (startDateChanged || dueDateChanged || endDateChanged) {
-			const startDateValue = typeof startDate !== 'undefined' ? startDate : this.startDate();
-			const dueDateValue = typeof dueDate !== 'undefined' ? dueDate : this.dueDate();
-			const endDateValue = typeof endDate !== 'undefined' ? endDate : this.endDate();
+			const startDateValue = this._getDateValue(startDate, this.startDate());
+			const dueDateValue = this._getDateValue(dueDate, this.dueDate());
+			const endDateValue = this._getDateValue(endDate, this.endDate());
 
 			const fields = [
 				{ name: 'startDate', value: startDateValue },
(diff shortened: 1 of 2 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -187,6 +187,30 @@ export class ActivityUsageEntity extends Entity {
...
+			action = datesEntity.getActionByName(Actions.activities.update);
+		}
+
+		if (!action) {
+			return;
+		}
+		const fields = [
+			{ name: 'startDate', value: startDate },
+			{ name: 'dueDate', value: dueDate },
+			{ name: 'endDate', value: endDate }
```
spandanabaruah: Here we are passing all three dates (startDate, dueDate and endDate) from the UI to the backend -- however, is this required to do? Would it be better to just pass the dates that have changed their values and use the backend to source the unchanged dates' values?

mdgbayly: Hmm, good question. My initial reaction is we should only send the dates that have changed. That tends to be how we have done other PATCH APIs. I think the only reason why we might want to always sends all dates, is if we want to guard against the possibility that someone is working with stale data e.g. someone else has modified the dates whilst the current user has the edit page open and there is the possibility that a different user has modified one of the dates to be inconsistent with the current user's change. Then the current user would get a validation error even though the dates they see on the page are consistent when they clicked save.

But if we allow them to send all their dates even though they have not changed some of them, that means they clobber other users changes even though they didn't explicitly change the field.

Might be a question for Joseph.

In general I think our goal is to not overwrite other changes if the current user didn't make changes, hence we only send changed data.  So my preference would be to not send all the dates and only send ones that have changed in the UI. But it might mean we have to make our validation handling smarter e.g. by refreshing the dates from the backend if we detect a validation error caused by stale data.

src/activities/ActivityUsageEntity.js
```diff
@@ -418,14 +442,20 @@ export class ActivityUsageEntity extends Entity {
...
-			activity.dueDate !== this.dueDate()) {
-			await this.setDueDate(activity.dueDate);
-		}
-
 		if (typeof activity.isDraft !== 'undefined' &&
 			activity.isDraft !== this.isDraft()) {
 			await this.setDraftStatus(activity.isDraft);
 		}
+
+		const startDateChanged = typeof activity.startDate !== 'undefined';
```
mdgbayly: Assuming we do decide to only send the changed dates, then I would probably move all this change tracking logic to the `setDates` implementation.

So here just call `this.setDates(activity.startDate, activity.endDate, activity.dueDate)`.

And in `setDates` do something like i did in the assignment implementation.

```
		if (typeof startDate !== 'undefined' && startDate !== this.startDate()) {
			fields.push({ name: 'startDate', value: startDate });
		}

       ...
       if (fields.length > 0) {
       // perform action
       }
```

src/activities/ActivityUsageEntity.js
emil-furniss: Confirmed with Joseph and apparently we _do_ want to overwrite saved changes made by any other user. It's what quizzing does as well when editing questions.
In that case we'll keep this as-is, so that the backend always has all the new date values to do the validation.

mdgbayly: I'd like to double check that because it means it's inconsistent with how we're doing other things. e.g. if the user edits the name, we don't save the instructions too.

Regardless, even if we stick with this approach of always saving all dates, I think we should enhance the `*DateChanged` checks to determine if the date has actually changed rather than just checking for undefined. If we leave it as is it means we are saving all the dates every time we save, which is again inconsistent with how we are doing other things?

Unless that is also what Joseph wants? To save everything every time? But that means we are making lots of extra API calls, and for some things it's harder to do and requires comparison checking on the backend. eg. for attachments.

emil-furniss: ^ Updated so it only saves when at least one of the dates has been changed.

PR #133, opened 2020-02-13
Kanderson/us112850 display rubrics associations

kieranderson:
Adding associations to be used to display rubric associations

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -428,4 +428,16 @@ export class ActivityUsageEntity extends Entity {
 			await this.setDraftStatus(activity.isDraft);
 		}
 	}
+
+	/**
+	 * @returns {string} The href to the associations collection.
+	 */
+	getRubricAssociationsHref(){
+		if (!this._entity) {
+			return [];
+		}
+
+		return this._entity.getLinkByRel(Rels.Activities.associations).href;
+
+	}
 }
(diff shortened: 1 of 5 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -485,4 +485,16 @@ export class ActivityUsageEntity extends Entity {
...
+
+	/**
+	 * @returns {string} The href to the associations collection.
+	 */
+	getRubricAssociationsHref(){
+		if (!this._entity) {
+			return [];
+		}
+
+		return this._entity.getLinkByRel(Rels.Activities.associations).href;
```
ppaskaris-d2l: Null-check this, just in case.

src/activities/Associations.js
```diff
@@ -0,0 +1,25 @@
...
+export class Associations extends Entity {
+
+	getSingleAssocationHrefs() {
+		if (!this._entity) {
+			return [];
+		}
+
+		const singleAssociations = this._entity.getSubEntitiesByClass(Classes.associations.singleAssociation);
+		const singleAssociationsHrefs = [];
+		singleAssociations.forEach(a => {
```
ppaskaris-d2l: Consider: `singleAssociations.map( a => a.getLinkByRel('self').href )` if you are going to do the unconditional `href` property access here (seems pretty safe for a `self` rel).

PR #135, opened 2020-02-24
Release conditions (Part 1)

ppaskaris-d2l:
[US109699](https://rally1.rallydev.com/#/detail/userstory/332018308528?fdp=true)
[US112831](https://rally1.rallydev.com/#/detail/userstory/360464559620?fdp=true)
[US112832](https://rally1.rallydev.com/#/detail/userstory/360464569172?fdp=true)
[US112859](https://rally1.rallydev.com/#/detail/userstory/360697030860?fdp=true)
[US113905](https://rally1.rallydev.com/#/detail/userstory/371431522876?fdp=true)

prereq for all of those ⬆️ 

This PR adds hypermedia for using legacy release conditions in the assignments editor. Since this is still quite expedient, I'm prefixing everything with `--legacy-` so that I don't steal any good names for this bridge version.

### DIFF
--- a/src/es6/Entity.js
+++ b/src/es6/Entity.js
@@ -43,7 +43,9 @@ export class Entity extends EntitySirenProperties {
 	 */
 	dispose() {
 		this._subEntities.forEach(entity => dispose(entity));
-		this._listener.remove();
+		if (this._listener) {
+			this._listener.remove();
+		}
 	}
 	/**
 	 * Protected: Add a listener to a subentity of this entity.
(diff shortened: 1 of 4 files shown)


src/es6/Entity.js
```diff
@@ -43,7 +43,9 @@ export class Entity extends EntitySirenProperties {
 	 */
 	dispose() {
 		this._subEntities.forEach(entity => dispose(entity));
-		this._listener.remove();
+		if (this._listener) {
```
m6jones, line 46: @kieranderson  Hey this being added probably means you should be using a selflessEntity instead of a full fledged entity. I am refreshing on the code and noticed this change and it a red flag for that.

PR #138, opened 2020-02-26
Added method for retrieving all completion type options

pureooze:
For [US112499](https://rally1.rallydev.com/#/110294864140d/detail/userstory/356264445024)

We need this method for MobX to allow validation of completion types when a user switches submission types. This method is very similar to `completionTypeOptions` but does not do the filtering for the currently selected submission type.


pureooze:
> cut-over first in the other repo before deleting the old one, I assume?

Yep, I want to move all usages of the old one over, then we can delete it 🗡

PR #139, opened 2020-02-27
add support for score and grade to activity usage aggregate save

mdgbayly:
[US113166](https://rally1.rallydev.com/#/29180338367d/detail/userstory/362987616496)

I've simplified the saving of the score and grade to just use the existing method/action that was being used to just set the score and add a new grade. There were existing methods which were more aligned to autosave which were geared around being very granular around whether we were setting score, adding to grades, removing from grades etc. But this would have made things much more complicated than necessary to handle the general save case. Also for the general save case we want to pass the `inGrades` flag regardless of whether it has changed because we want the latest state to always reflect what the user was seeing.

Also added tests to prevent updating grades if cannot edit grades.

Also, had to fix up some of the existing dates tests.
The ActivityUsage tests were actually missing from the `index.html` file so a lot of them were not actually working as intended. A reminder again to verify that your tests fail correctly so that you can check that they are actually being run. Also, you can run the tests in the browser using `npm run serve` to verify that your tests are running and to debug them.

### DIFF
--- a/test/activities/ActivityUsageEntity.js
+++ b/test/activities/ActivityUsageEntity.js
@@ -5,11 +5,14 @@ import { testData } from './data/ActivityUsageEntity.js';
 import { getFormData } from '../utility/test-helpers.js';
 
 describe('ActivityUsageEntity', () => {
-	let entity, readonlyEntity, entityJson;
+	let entity, entityCannotEditGrades, readonlyEntity, entityJson, entityJsonCannotEditGrades;
 
 	beforeEach(() => {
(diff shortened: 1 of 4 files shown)


test/activities/ActivityUsageEntity.js
```diff
@@ -298,6 +301,102 @@ describe('ActivityUsageEntity', () => {
 				expect(fetchMock.done());
 			});
 		});
+
+		describe('score and grade', () => {
```
wongvincent: nice tests!

PR #142, opened 2020-03-03
Saving assignment type from MobX

pureooze:


src/activities/assignments/AssignmentEntity.js
```diff
@@ -610,6 +614,22 @@ export class AssignmentEntity extends Entity {
 			fields.push({ name: 'completionType', value: assignment.completionType });
 		}
 
+		if (typeof assignment.isIndividualAssignmentType !== 'undefined') {
+			if (assignment.isIndividualAssignmentType !== this.isIndividualAssignmentType()) {
+				fields.push({ name: 'isIndividualAssignmentType', value: assignment.isIndividualAssignmentType });
+			}
+
+			if (!assignment.isIndividualAssignmentType && !this.isGroupAssignmentTypeDisabled()) {
+				fields.push({ name: 'groupTypeId', value: assignment.groupTypeId });
```
pureooze: See https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/assignments/AssignmentEntity.js#L306-L309

src/activities/assignments/AssignmentEntity.js
```diff
@@ -610,6 +614,22 @@ export class AssignmentEntity extends Entity {
 			fields.push({ name: 'completionType', value: assignment.completionType });
 		}
 
+		if (typeof assignment.isIndividualAssignmentType !== 'undefined') {
+			if (assignment.isIndividualAssignmentType !== this.isIndividualAssignmentType()) {
+				fields.push({ name: 'isIndividualAssignmentType', value: assignment.isIndividualAssignmentType });
```
ppaskaris-d2l: This field is not present on the [controller definition](https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Controllers/FoldersController.cs?r=65c56194#215). What are you trying to do here?

ppaskaris-d2l: This parameter is not present on [the controller](https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Controllers/FoldersController.cs?r=65c56194#203. What are you trying to do here?

_I did not mean to delete the previous version of this comment._

pureooze: Oh good catch! 🤦‍♂ I think we could just remove this if statement? Wouldn't the section below this be sufficient at setting the folder type?

src/activities/assignments/AssignmentEntity.js
```diff
@@ -610,6 +614,22 @@ export class AssignmentEntity extends Entity {
...
+			}
+
+			if (!assignment.isIndividualAssignmentType && !this.isGroupAssignmentTypeDisabled()) {
+				fields.push({ name: 'groupTypeId', value: assignment.groupTypeId });
+				fields.push({ name: 'folderType', value: 1 });
+			} else {
+				const action = this._getActionSetToIndividualAssignmentType();
```
ppaskaris-d2l: I _super_ do not like this. I would prefer you just hard-code `folderType = 2` and `groupTypeId = null`

src/activities/assignments/AssignmentEntity.js
ppaskaris-d2l: I think so, yeah.

pureooze:
> okiedokie, the double negatives are confusing though

Hmm thats a really good point, I'm going to make a separate PR to fix that. Just want to merge this so we can get the UI stuff reviewed!

PR #144, opened 2020-03-04
Vwong/us111000 render grade candidates

wongvincent:
So we now have
`GradeCandidateCollectionEntity` which contains a list of `GradeCandidateEntity`s.
`GradeEntity` is what is returned by the `/grades/cache-primer`.
`GradeCategoryEntity` is what is returned by the `/grade-categories/cache-primer`.

A `GradeEntity` and `GradeCategoryEntity` will contain the properties of the item such as its name.
The `GradeCandidateEntity` will contain the associate action. (Activity specific)

### DIFF
--- a/test/activities/GradeCandidateEntity.js
+++ b/test/activities/GradeCandidateEntity.js
@@ -1,25 +1,20 @@
+/* global fetchMock */
+
+import { getFormData } from '../utility/test-helpers.js';
 import { GradeCandidateEntity } from '../../src/activities/GradeCandidateEntity.js';
 import { testData } from './data/GradeCandidateEntity.js';
 
 describe('GradeCandidateEntity', () => {
-	let sandbox;
-
-	beforeEach(() => {
-		sandbox = sinon.sandbox.create();
-	});
-
 	afterEach(() => {
-		sandbox.restore();
+		fetchMock.reset();
 	});
 
 	describe('Grade', () => {
-		let entity;
-		let associateGradeSpy;
+		let entity, entityJson;
 
 		beforeEach(() => {
-			const entityJson = window.D2L.Hypermedia.Siren.Parse(testData.gradeCandidateEntity.grade);
+			entityJson = window.D2L.Hypermedia.Siren.Parse(testData.gradeCandidateEntity.grade);
 			entity = new GradeCandidateEntity(entityJson);
-			associateGradeSpy = sandbox.spy(entity, 'associateGrade');
 		});
 
 		it('gets href', () => {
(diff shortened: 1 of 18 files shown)


test/activities/GradeCandidateEntity.js
```diff
@@ -41,4 +43,70 @@ describe('GradeCandidateCollectionEntity', () => {
 			expect(associateGradeSpy.returnValues[0]).to.be.a('promise');
```
mdgbayly: I know you didn't write this, but I don't think this test is actually doing anything, and doesn't really need to use a spy.
All it does is verifies that the `associateGrade` returns a promise, which by itself is not particularly useful, and it could have been verified by just checking the return value of `associateGrade`. Ideally tests like this would be using fetchMock to verify that the right API is called with the correct parameters.

test/activities/GradeCandidateEntity.js
```diff
@@ -41,4 +43,70 @@ describe('GradeCandidateCollectionEntity', () => {
...
+			expect(entity.isCategory()).to.be.false;
+		});
+
+		it('can associate grade', () => {
+			expect(entity.canAssociateGrade()).to.be.false;
+		});
+
+		it('cannot associate to a grade', () => {
+			entity.associateGrade();
+			expect(associateGradeSpy.returnValues).to.not.be.a('promise');
```
mdgbayly: ditto - i don't think this needs to use a spy?

test/activities/GradeCandidateEntity.js
```diff
@@ -41,4 +43,70 @@ describe('GradeCandidateCollectionEntity', () => {
...
+			const expected = entity.getGradeCandidates();
+			expect(expected).to.be.an('array');
+			expect(expected).to.have.lengthOf(1);
+		});
+
+		it('is a category', () => {
+			expect(entity.isCategory()).to.be.true;
+		});
+
+		it('can associate grade', () => {
```
mdgbayly: should be `can_not_ associate grade`?

test/activities/GradeCandidateEntity.js
mdgbayly: Also, this will actually try and make an API calls which can screw up other tests that are trying to verify API calls using mocks.

test/activities/GradeEntity.js
```diff
@@ -0,0 +1,67 @@
...
+		it('gets baseWeight', () => {
+			expect(entity.baseWeight()).to.equal(10);
+		});
+
+		it('gets maxPoints', () => {
+			expect(entity.maxPoints()).to.equal(100);
+		});
+	});
+
+	describe('Grade Category', () => {
```
mdgbayly: Is there a reason we're also testing grade categories with the GradeEntity given that we now have a distinct GradeCategoryEntity?

wongvincent: oops...

test/activities/GradeCandidateEntity.js
wongvincent: oops

test/activities/GradeCandidateEntity.js
wongvincent: fixed

PR #147, opened 2020-03-06
Adding association collection and more to association

kieranderson:


src/activities/Association.js
```diff
@@ -1,12 +1,11 @@
...
-import { Actions, Rels } from '../hypermedia-constants';
+import { Actions, Classes, Rels } from '../hypermedia-constants';
 import { performSirenAction } from '../es6/SirenAction';
-
 /**
  * Association class representation of an association.
  */
 
-export class Association extends Entity {
+export class AssociationEntity extends Entity {
```
ppaskaris-d2l: Can you also `export const Association = AssociationEntity;` so it doesn't break? multi-repo madness and all...

src/activities/AssociationCollectionEntity.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+export class AssociationCollectionEntity extends Entity {
+
+	getAllAssociations() {
+		if (!this._entity) {
+			return [];
+		}
+
+		const associations = this._entity.getSubEntitiesByClass(
```
ppaskaris-d2l: Strongly consider checking the `rel`

src/activities/Associations.js
```diff
@@ -1,26 +1,19 @@
...
 
-		const singleAssociations = this._entity.getSubEntitiesByClass(
-			Classes.associations.singleAssociation
-		);
-		const singleAssociationsHrefs = singleAssociations.map(
-			a => a.getLinkByRel('self').href
-		);
-
-		return singleAssociationsHrefs;
+		return this._entity.getSubEntitiesByRel('item');
```
ppaskaris-d2l: Won't this just return vanilla siren entities, and not the wrapper objects? Is that what you want?

Also, now no filtering by single/potential association?

I guess the intent of this method has now changed.

kieranderson: I had a separate CollectionEntity that was using this but I moved it here to associations

PR #150, opened 2020-03-10
add assignment delete

stevenao:


mdgbayly:
@stevenao - this is a pretty simple method, but it wouldn't hurt to add a simple unit test using something like fetchMock.

https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/test/activities/assignments/AssignmentEntity.js#L64

Less concerned about this because it doesn't really do much other than execute the action which is really only going to be testing the `SirenAction` stuff, but it feels like a good habit to get into.

PR #152, opened 2020-03-10
Fix bug where changing group type would not save

pureooze:


src/activities/assignments/AssignmentEntity.js
```diff
@@ -610,14 +610,13 @@ export class AssignmentEntity extends Entity {
 			fields.push({ name: 'completionType', value: assignment.completionType });
 		}
 
-		if (typeof assignment.isIndividualAssignmentType !== 'undefined' &&
-				assignment.isIndividualAssignmentType !== this.isIndividualAssignmentType()) {
```
ppaskaris-d2l: Can you please explain why this was broken and how your edits fix it?

pureooze: When we load the assignment type as a group type and only change the group selected (the `groupTypeId`) the `isIndividualAssignmentType` value doesnt change and so the change is ignored.

ppaskaris-d2l: Oh. Would be nice to have the change-tracking though, but I don't know how much it actually matters. Perhaps you could check:

1. Did the assignment type change
2. If it's a group assignment type, did the group category change

PR #157, opened 2020-03-16
Removed the completionTypeOptions method from AssignmentEntity

pureooze:
In #138 I added a method to retrieve all the completion types. Just removing the old method now since its not used.

Usages: https://search.d2l.dev/search?project=GitHub&full=completionTypeOptions&defs=&refs=&path=&hist=&type=&si=full


docs/AssignmentEntity.html
```diff
@@ -784,101 +784,8 @@ <h5>Returns:</h5>
 
     
 
-
```
kieranderson: Did you manually remove this or generated the docs and this is what it removed?

pureooze: I removed it manually. I'm not sure how these are generated, there was no documentation for them in the README 😢

pureooze: Hmm so running that command seems to regen the docs but it updates >20 files so I think we can do this is a separate PR? Really feel like this should just be a step in the build...

kieranderson: Yeah if it hasn't been updated in a while it can be large. Just wasn't sure if there was any links that point to what you've deleted. Probably not though.

PR #163, opened 2020-03-20
Vwong/us114169 associate new grade

wongvincent:
**Call `associate-new-grade` action**
This will send a PATCH request to the `activity-usage` endpoint. The start/due/end dates also use this endpoint.
In a future PR, the two PATCH requests will be combined into one.

**Fix hidden fields appended twice when the original fields argument is not `undefined`**
When `fields` is passed in as a parameter, because it is an array, appending hidden fields in the `_getEntityUrl` function and then again in the `_performSirenAction` function modifies the original `fields` variable. This results in  the hidden field to be added twice.

### DIFF
--- a/src/es6/SirenAction.js
+++ b/src/es6/SirenAction.js
@@ -30,14 +30,14 @@ const _getEntityUrl = function(action, fields) {
 	}
 
 	let url = new URL(action.href, window.location.origin);
-
-	if (fields) {
-		fields = _appendHiddenFields(action, fields);
+	let tempFields = fields ? [...fields] : [];
+	if (tempFields.length) {
+		_appendHiddenFields(action, tempFields);
 	} else {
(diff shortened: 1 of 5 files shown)


src/es6/SirenAction.js
```diff
@@ -116,7 +116,7 @@ const _performSirenAction = function(action, fields, tokenValue) {
 	let body;
 
 	if (fields) {
-		fields = _appendHiddenFields(action, fields);
+		_appendHiddenFields(action, fields);
```
emil-furniss: Might it be better to remove all the field-setting logic from `_getEntityUrl`, so it just uses the fields passed in directly, and then move the `const url` assignment to below line 122?
If I read it correctly, at that point `fields` should contains all fields, both hidden and not, and avoids duplicating logic and work just to avoid some side-effects.

PR #164, opened 2020-03-25
Call activityusage save with combined patch calls

wongvincent:
Create a new action that combine fields from actions with the same href and method into one action, then calls it.
The change combine the dates and the associate-new-grade PATCH requests to activity-usage into one action/request.

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -218,7 +218,7 @@ export class ActivityUsageEntity extends Entity {
 		if (!dates) return;
 		if (!this._hasDatesChanged(dates.startDate, dates.dueDate, dates.endDate)) return;
 
-		const datesActionAndFields = this._generateDatesAction(dates.startDate, dates.dueDate, dates.dueDate, true);
+		const datesActionAndFields = this._generateDatesAction(dates.startDate, dates.dueDate, dates.endDate, true);
 		if (datesActionAndFields) {
 			await performSirenAction(this._token, datesActionAndFields.action, datesActionAndFields.fields);
 		}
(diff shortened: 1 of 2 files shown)


src/es6/SirenAction.js
```diff
@@ -151,6 +151,68 @@ const _performSirenAction = function(action, fields, tokenValue) {
...
+			fields = _getSirenFields(actionAndField.action);
+		}
+		fields.forEach(field => fieldsMap.set(field.name, field.value));
+
+		methodsMap.set(method, fieldsMap);
+		hrefsMap.set(href, methodsMap);
+	});
+
+	const combinedActions = [];
+	for (const [href, methodsMap] of hrefsMap.entries()) {
```
mdgbayly: This could probably be built in place to avoid the second loop, but OK to leave for now.
If we get some unit tests in a follow up PR :-) then it should be easy to refactor this kind of thing.

mdgbayly: I guess for the fields at least though, using the map in the first loop makes it easier to manage (ignore?/overwrite?) duplicate field names. As you are translating the mapped fields into objects in the second loop.

mdgbayly: Currently we are losing any other field information in this process, e.g. `type` but the only place we currently seem to use that is in the `_performSirenAction._addHiddenFields` which I think will be called again on the aggregated action, only there will never be any hidden fields which is kind of by definition at that point. But something for us to be aware of for future changes maybe.

src/activities/ActivityUsageEntity.js
```diff
@@ -533,19 +546,19 @@ export class ActivityUsageEntity extends Entity {
 		const createNewGradeItem = scoreAndGrade.inGrades && !associatedGrade && scoreAndGrade.associateNewGradeAction;
 		const associateToExistingGrade = scoreAndGrade.inGrades && associatedGrade && associatedGrade.canAssociateGrade() && this.gradeHref() !== associatedGrade.href();
 		if (createNewGradeItem) {
```
mdgbayly: Definitely having a hard time getting my head around what is going on with all the grade and score stuff, but no surprise there. Hopefully if we can further rationalize how we are handling scores and grades and also add them to the activity usage in the future, we will be able to simplify this too.

wongvincent:
unit tests 🎉  https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/166 🎉

PR #170, opened 2020-03-27
Omsmith/org image use link

omsmith:


src/es6/Entity.js
```diff
@@ -103,27 +103,38 @@ export class Entity extends EntitySirenProperties {
...
 		// Clean up if that href has already been added.
-		if (this._subEntities.has(href)) {
-			dispose(this._subEntities.get(href));
+		if (this._subEntities.has(link.href)) {
+			dispose(this._subEntities.get(link.href));
 		}
 		this._subEntitiesLoadStatus.push(new Promise((resolve) => {
-			entityFactory(entityType, href, this._token, (entity) => {
-				this._subEntities.set(href, entity);
+			entityFactory(entityType, link, this._token, (entity) => {
```
omsmith: Support added in d2l-polymer-siren-behaviors@1.4.0

omsmith:
Have tests running locally now, will sort through what updates need to happen there.

src/es6/Entity.js
jopnick: shouldn't this be `link.href`?

omsmith: Nope. It's passing the Link object to polymer-siren-behaviors so they the `rel` on it can be inspected

omsmith:
Had to delete Travis' cache to get tests to pass due to the polymer-siren-behaviors version bump (because it isn't an explicit bump here)

PR #171, opened 2020-03-30
DE38379 Added assignmentHasSubmissions method to AssignmentEntity.js

pureooze:
For [DE38379](https://rally1.rallydev.com/#/110294864140d/detail/defect/377932041916), added a method to retrieve if an assignment has submissions. 

## Why?
At the moment the Hypermedia API will never return a `folder-type` entity with the `has-submissions` class because it does not exist in the API. The plan is to rename the `read-only` class when the user HAS permission to edit the entity to `has-submissions` since that is what it was really being used for. Then when the user does NOT have permissions the API will actually respond with `read-only`.

## What is next?
The plan is as follows:
1. add `assignmentHasSubmissions` method to siren-sdk
2. add property in assignment MobX store to retrieve this value along with `read-only`
3. replace the `read-only` class in the hypermedia API with `has-submissions` when the user HAS permission to edit
4. update the `isAssignmentTypeReadOnly` method in siren-sdk to rely on the presence of actions
5. update `save` method in siren-sdk to also check if the assignment has submissions
6. remove `read-only` class in the hypermedia API since we don't need it
7. verify UI to make sure permissions reflected as expected

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -202,6 +202,22 @@ export class AssignmentEntity extends Entity {
 	}
 
 	/**
+	 * @returns {bool} If the assignment has submissions
+	 */
+	assignmentHasSubmissions() {
+		if (!this._entity) {
+			return false;
+		}
+
+		const subEntity = this._entity.getSubEntityByRel(Rels.Assignments.folderType);
+		if (!subEntity) {
+			return false;
+		}
+
+		return subEntity.hasClass(Classes.assignments.assignmentType.hasSubmissions);
+	}
+
+	/**
 	 * @returns {bool} If the assignment type "group assignment" can be set
 	 */
 	isGroupAssignmentTypeDisabled() {
--- a/src/hypermedia-constants.js
+++ b/src/hypermedia-constants.js
@@ -219,7 +219,8 @@ export const Classes = {
 		assignmentType: {
 			individual: 'individual',
 			noGroupType: 'no-group-type',
-			readOnly: 'read-only'
+			readOnly: 'read-only',
+			hasSubmissions: 'has-submissions'
 		},
 		attachment: 'attachment',
 		attachmentList: 'attachment-list',


ppaskaris-d2l:
If we don't have any actions to change the type, do we _need_ the `read-only` class, or is it redundant and can be removed? I understand the need for `has-submissions`, so you can put a string saying "this assignment has submissions, that's why you can't edit it even tho you normally could."

Perhaps it could just be:

1. Have permission, no submissions => actions present on the entity
2. Have permission, submissions => no actions present, `has-submissions` class <-- "read-only"
3. No permissions, no submissions => no actions, no `has-submissions` class <-- "read-only"
4. No permissions, has submissions => no actions, no `has-submissions` class <-- "read-only"

What am I missing in my understanding?

src/activities/assignments/AssignmentEntity.js
```diff
@@ -201,6 +201,22 @@ export class AssignmentEntity extends Entity {
...
+	 * @returns {bool} If the assignment has submissions
+	 */
+	assignmentHasSubmissions() {
+		if (!this._entity) {
+			return false;
+		}
+
+		const subEntity = this._entity.getSubEntityByRel(Rels.Assignments.folderType);
+		if (!subEntity) {
+			return false;
```
ppaskaris-d2l: Aside: this is a case where `undefined` seems appropriate - the absence of a value! It might very well have submissions!

Don't actually change it though, just an observation.

pureooze: Hmm should we maybe actually default to true? Since if we don't know if an assignment has submissions we may attempt to change its properties when it actually could have submissions?

ppaskaris-d2l: The server is the _authority_ on whether you can take an action. If it provides an action, assume that you can do it. Otherwise, assume that you can't. Inferring this information through classes, etc. is what I'm raising in my top-level comment above.

In short: if action, can do, otherwise, no can do! Let's not rely on classes for things built-in to Siren.

pureooze:
Github doesn't allow replying directly to a comment 🤦‍♂ so I'm just going to do this here

So the issue is that currently the [UI does not rely on the presence of an action to determine if the assignment type is editable or not](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-type-editor.js#L131).

But now that I think about it after reading your comment, we could just change `isAssignmentTypeReadOnly` to check if the `set-to-individual` and `set-to-group` actions are present. My understanding is that if the assignment can be modified, both are present and if they are not then it must mean we cannot update the assignment type.

ppaskaris-d2l:
I can imagine a world where you can change which group an assignment belongs to without be able to switch it to individual, but yeah, general rule I would reckon is: no actions whatsoever, no edits whatsoever.

pureooze:
I updated the plan in my first comment to reflect this discussion 👍

ppaskaris-d2l:
> * update `save` method in siren-sdk to also check if the assignment has submissions

Why? If there are submissions, then _the actions won't be present_ and you _cannot submit them_.

pureooze:
>Why? If there are submissions, then the actions won't be present and you cannot submit them.

Ah yeah good point!

PR #172, opened 2020-03-31
add files per submission limit

stevenao:


src/activities/assignments/AssignmentEntity.js
```diff
@@ -489,6 +489,41 @@ export class AssignmentEntity extends Entity {
...
+		}
+
+		const fieldValue = action.getFieldByName('filesSubmissionLimit').value.find(v => {
+			return v.value === filesSubmissionLimit;
+		});
+		if (!fieldValue) {
+			return;
+		}
+
+		const filesSubmissionLimitMap = { 'unlimited': 0, 'one': 1 };
```
mdgbayly: This seems a bit non-standard. Ideally we shouldn't have to convert between the value representations in the original action and some other value for the actual execution of the action. This is baking out-of-band knowledge into the client that should really only be made available through the hypermedia API. I realize this was done based on my comments to make the values in the API clearer. But I think if we have to to either go all the way and allow the string values to be passed as part of the API call and do the translation in the controller or stick to string representations of the numeric values.

Does the `PostParamBinder` in the API support translating directly from string representations of enums to the strongly typed enum by any chance? eg. so that you could use the values `Unlimited` and `OneFilePerSubmission`?

stevenao: Just tested it, it does what you suggested. I will make another LMS pr.

PR #173, opened 2020-03-31
DE38379 Update siren-sdk to allow displaying info even when in read only mode

pureooze:
Continuing work from https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/171

This PR makes the `isAssignmentTypeReadOnly` method check the entity for the assignment type actions rather than relying on the `read-only` class.
Also return the submission type even if the assignment is read only.

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -194,7 +194,7 @@ export class AssignmentEntity extends Entity {
 		if (!this._entity) {
 			return false;
 		}
-		return !this._entity.hasActionByName(Actions.assignments.setToGroup) && 
+		return !this._entity.hasActionByName(Actions.assignments.setToGroup) &&
 			!this._entity.hasActionByName(Actions.assignments.setToIndividual) ;
 	}
 
@@ -363,7 +363,7 @@ export class AssignmentEntity extends Entity {
 
 	_getReadOnlySubmissionTypeOptions() {
 		if (!this._entity) {
-			return []
+			return [];
 		}
 
 		const currentSubmissionType = this._entity.properties.submissionType;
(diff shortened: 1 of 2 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -379,13 +385,17 @@ export class AssignmentEntity extends Entity {
...
+		if (!this._entity) {
 			return [];
 		}
 
 		const action = this._entity.getActionByName(Actions.assignments.updateSubmissionType);
+		if (!action) {
+			return this._getReadOnlySubmissionTypeOptions();
+		}
+
 		if (!action.hasFieldByName('submissionType')) {
```
kieranderson, line 397: Is it possible for this action to exist without that field? Or is this just a safety check.

pureooze: Looks like this is the only place its used, and it has the submission type field (but this is also a safety check) https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs?r=37b0acf5#1040

src/activities/assignments/AssignmentEntity.js
```diff
@@ -194,11 +194,8 @@ export class AssignmentEntity extends Entity {
 		if (!this._entity) {
 			return false;
 		}
-		const subEntity = this._entity.getSubEntityByRel(Rels.Assignments.folderType);
-		if (!subEntity) {
-			return false;
-		}
-		return subEntity.hasClass(Classes.assignments.assignmentType.readOnly);
+		return !this._entity.hasActionByName(Actions.assignments.setToGroup) &&
```
kieranderson, line 197: These actions got moved to the main assignment entity?

pureooze: Nope that is my bad, pushing a fix

PR #177, opened 2020-04-07
Yup, Fix edge.

m6jones:
Using Edge (42.17134.1098.0) these actions would never be performed. 

Now they are!

https://rally1.rallydev.com/#/357252966636d/detail/defect/365347773396

### DIFF
--- a/src/organizations/OrganizationCollectionEntity.js
+++ b/src/organizations/OrganizationCollectionEntity.js
@@ -113,7 +113,8 @@ export class OrganizationCollectionEntity extends Entity {
 		const entity = await performSirenAction(
 			this._token,
 			searchAction,
-			fields
+			fields,
+			true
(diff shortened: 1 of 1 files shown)


zinabat:
I am legitimately flabbergasted that this is the fix.

maboelsoudd2l:
does this coincidentally fix the "create learning path"?

src/organizations/OrganizationCollectionEntity.js
```diff
@@ -133,7 +134,8 @@ export class OrganizationCollectionEntity extends Entity {
 		const entity = await performSirenAction(
 			this._token,
 			createOrgUnitAction,
-			fields
+			fields,
+			true
```
m6jones, line 138: This one fixes the creation.

PR #178, opened 2020-04-09
Added ActivityGradeEntity, tests and updated docs

dmackenz:
This new Entity is for the new `d2l-grade-result` component that is being used in `consistent-evaluation`.


src/activities/ActivityGradeEntity.js
```diff
@@ -0,0 +1,41 @@
...
+export class ActivityGradeEntity extends Entity {
+	/**
+	 * @returns {('Numeric'|'LetterGrade')} The Grade's score type.
+	 */
+	scoreType() {
+		return this._entity && this._entity.properties && this._entity.properties.scoreType;
+	}
+
+	/**
+	 * @returns {number|null} The score given for a Numeric grade (numerator).
```
m6jones, line 15: undefined is also something that can be returned. Not that important though.

dmackenz:
@AlexBedley @rowansilver Just wanted to let you both know that this is a thing now

PR #179, opened 2020-04-13
US115409 Support alignments collection entity

emil-furniss:
This adds a new class for the alignments collection entity with a couple of basic methods, and a getter for the alignments href on the activity usage entity. These will be used by `d2l-activities` to show/hide the correct learning outcomes UI in FACE.


wongvincent:
I don't have the big picture of what's happening, but this looks fine to me!

PR #181, opened 2020-04-17
Add discover link

edwardg8:


src/organizations/OrganizationCollectionEntity.js
```diff
@@ -26,6 +27,13 @@ export class OrganizationCollectionEntity extends Entity {
 		return this._entity.getSubEntitiesByRel(rels.organization);
 	}
 
+	discoverCourses() {
```
m6jones, line 30: This should be named discoverActivities or just Activities.

edwardg8: I'll change it to activities, isn't `activity` completely different entity from `organization`

BhamelinD2L: @m6jones Would it possibly be better to create a new `DiscoverCollectionEntity` in the actions folder? I think our API exactly matches this one aside from the `rel` call otherwise, which is why I suggested we use it to Ed in the first place.

@edwardg8 our sub-entities are of type `activity`, which is why they plug in to the `d2l-activity-card` in the `d2l-activity-list`. The discover homepage currently fetches them and then feeds them into the card list without needing to care about their type or content. You can get the `organization` from the `activity` very easily, which is what Search does to populate the list items, and I think discovery-course likely does too.
I'm not sure if any of the existing Activity APIs would fit our activities exactly. The search list and course summary could both benefit from that, but I'd worry about it another time.

edwardg8: having a separate CollectionEntity also sounds good, I'm ok with either

edwardg8: @BhamelinD2L I think it's better to proceed with the changes, so my other PR is not completely blocked. If we need more routes from Discover later, we can create its own Entity or whenever that is needed.

PR #182, opened 2020-04-24
Add deferred add/remove association actions

awikkerink:
* ~~API work being done by @jopnick~~

### DIFF
--- a/src/activities/Association.js
+++ b/src/activities/Association.js
@@ -1,10 +1,10 @@
 import { Entity } from '../es6/Entity.js';
 import { Actions, Classes, Rels } from '../hypermedia-constants';
 import { performSirenAction } from '../es6/SirenAction';
+
 /**
  * Association class representation of an association.
  */
-
 export class AssociationEntity extends Entity {
 
 	/**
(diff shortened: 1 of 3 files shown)


src/activities/Association.js
```diff
@@ -33,24 +33,26 @@ export class AssociationEntity extends Entity {
 	/**
 	 * Deletes the association
 	 */
-	async deleteAssociation() {
-		if (!this._entity || !this._entity.hasActionByName(Actions.associations.deleteAssociation)) {
+	async deleteAssociation({ defer = false }) {
```
kieranderson, line 36: Are there more options being added to the params? Updating to `deleteAssociation( defer = false)` won't require a  change from the caller.

awikkerink: Yes, this is by design.

The usages will just need to set `defer = true`, then call `applyDeferredAssociations` to apply the changes.

kieranderson: Ok. Will need to make sure we update `d2l-activities` since this will break those `delete/createAssociation` calls if the `siren-sdk` version gets updated.

awikkerink: Why?

PR #187, opened 2020-05-07
US115945 Support uneval count property in competencies entity

emil-furniss:


test/competencies/CompetenciesEntity.js
```diff
@@ -13,6 +13,10 @@ describe('CompetenciesEntity', () => {
 		expect(entity.associatedCount()).to.equal(123);
 	});
 
+	it('has unevaluated count', () => {
+		expect(entity.associatedCount()).to.equal(54);
```
wongvincent, line 17: `entity.unevaluatedCount`?

emil-furniss: shoot, yes! The entity was also wrong, and the build passed because this suite wasn't included in the `test/index.html` file so they weren't being run at all.

PR #188, opened 2020-05-12
Add methods for checking if association actions exist

kieranderson:


src/activities/Association.js
```diff
@@ -30,6 +30,10 @@ export class AssociationEntity extends Entity {
 		return this._entity.hasClass(Classes.associations.singleAssociation);
 	}
 
+	canDeleteAssociation() {
+		return this._entity && this._entity.hasActionByName(Actions.associations.deleteAssociation);
```
pureooze, line 34: If entity is undefined, this would return undefined. I noticed that in some methods like `isSingleAssociation` and some in the Assignments entity we return false for these cases. Would we want to do the same here?

PR #199, opened 2020-07-16
US109625 Default scoring rubric

pureooze:
This PR adds retrieval and saving of the default scoring rubric for an assignment.
This is not the best solution but due to the time constraint we are going with the solution and will come back and clean it up with the deferred rubric save work.

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -799,9 +799,9 @@ export class AssignmentEntity extends Entity {
 		}
 
 		if (typeof assignment.defaultScoringRubricId !== undefined && 
-			assignment.defaultGroupTypeId !== this.getDefaultScoringRubric() &&
+			assignment.defaultScoringRubricId !== this.getDefaultScoringRubric() &&
 			this.canEditDefaultScoringRubric()) {
-				fields.push({ name: 'defaultScoringRubricId_DoNotUse', value: assignment.defaultScoringRubricId });
+				fields.push({ name: 'defaultScoringRubricId', value: assignment.defaultScoringRubricId });
 		}
 
 		if (fields.length > 0) {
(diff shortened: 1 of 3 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -780,6 +798,12 @@ export class AssignmentEntity extends Entity {
 			}
 		}
 
+		if (typeof assignment.defaultScoringRubricId !== undefined &&
```
astadelbauer, line 801: should undefined be 'undefined' like above?

pureooze: Ah good catch, yes it should

src/hypermedia-constants.js
```diff
@@ -242,7 +242,8 @@ export const Classes = {
 	},
 	associations: {
 		singleAssociation: 'single-association',
-		potentialAssociation: 'potential-association'
+		potentialAssociation: 'potential-association',
+		defaultScoringRubricAssociation: 'default-scoring-rubric-association'
```
kieranderson, line 246: This class still going to be used?

pureooze: Ah no this was for the other way I was attempting yesterday, will remove.

src/activities/assignments/AssignmentEntity.js
```diff
@@ -780,6 +798,12 @@ export class AssignmentEntity extends Entity {
 			}
 		}
 
+		if (typeof assignment.defaultScoringRubricId !== 'undefined' &&
+			assignment.defaultScoringRubricId !== this.getDefaultScoringRubric() &&
+			this.canEditDefaultScoringRubric()) {
+			fields.push({ name: 'defaultScoringRubricId', value: assignment.defaultScoringRubricId });
```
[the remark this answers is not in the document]
pureooze: So in the assignment store I set it to a default value if there is none provided when we retrieve the id, I don't set it to a default when saving because the server does accept `null` as a value for this field.

PR #200, opened 2020-07-16
US109625 default scoring rubric

pureooze:
Bugfix for when there is nothing in the DB (new assignment)

(ignore the noise commits, I used the same branch by accident, will squash and merge)

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -799,9 +799,9 @@ export class AssignmentEntity extends Entity {
 		}
 
 		if (typeof assignment.defaultScoringRubricId !== undefined && 
-			assignment.defaultGroupTypeId !== this.getDefaultScoringRubric() &&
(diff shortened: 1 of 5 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -188,7 +188,13 @@ export class AssignmentEntity extends Entity {
 			return null;
 		}
 
-		return this._entity.properties.defaultScoringRubricId;
+		const defaultScoringRubricId = this._entity.properties.defaultScoringRubricId;
+
+		if (defaultScoringRubricId === undefined || defaultScoringRubricId === null) {
```
pureooze, line 193: Might be possible that ID is 0? 🤷

astadelbauer: If it's just rubrics, doesn't look like 0 is actually valid. Looks like it starts at 1: https://search.d2l.dev/xref/lms/lp/_dbschema/D2L.LP.Rubrics/Tables/RUBRICS.sql?r=55de0cab#5

PR #215, opened 2020-09-08
Setup basic content entity

imcdonald-d2l:
Setting up basic content entity in `siren-sdk` so activities repo can initialize a content store with this entity type. Basing this off a similar approach assignments took when setting up there `assignments-store.js` [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment-store.js#L7).  Assignments initializes an entity in their state and bases it off the entity definition in this repo. The use of it in the activities repo can be seen [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-assignment-editor/state/assignment.js#L32). 

This is just the initialization of the entity with a basic name property so I can setup a `content` store in activities. I assume this is the structure it will take but can easily adjust if it is not.

### DIFF
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -8,6 +8,6 @@ export class ContentEntity extends Entity {
 	 * @returns {string} Name of the content item
 	 */
 	name() {
-		return this._entity && this._entity.properties && this._entity.properties.name;
+		return this._entity && this._entity.properties && this._entity.properties.title;
 	}
 }
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -7,7 +7,7 @@ export class ContentEntity extends Entity {
 	/**
 	 * @returns {string} Name of the content item
 	 */
-	name() {
+	title() {
 		return this._entity && this._entity.properties && this._entity.properties.title;
 	}
 }
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -5,7 +5,7 @@ import { Entity } from '../../es6/Entity';
  */
 export class ContentEntity extends Entity {
 	/**
-	 * @returns {string} Name of the assignment
+	 * @returns {string} Name of the content item
 	 */
 	name() {
 		return this._entity && this._entity.properties && this._entity.properties.name;
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -1,0 +1,13 @@
+import { Entity } from '../../es6/Entity';
+
+/**
+ * ContentEntity class representation of a d2l content entity.
+ */
+export class ContentEntity extends Entity {
+	/**
+	 * @returns {string} Name of the assignment
+	 */
+	name() {
+		return this._entity && this._entity.properties && this._entity.properties.name;
+	}
+}
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -5,7 +5,7 @@ import { Entity } from '../../es6/Entity';
  */
 export class ContentEntity extends Entity {
 	/**
-	 * @returns {string} Name of the content item
+	 * @returns {string} Title of the content item
 	 */
 	title() {
 		return this._entity && this._entity.properties && this._entity.properties.title;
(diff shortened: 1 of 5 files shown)


test/activities/content/data/EditableContent.js
```diff
@@ -0,0 +1,10 @@
+export const editableContent = {
+	'class': [
+		'content'
+	],
+	'properties': {
+		'name': 'Extra Special Content'
```
imcdonald-d2l, line 6: @CodeBaboon @ChrisKraljevicD2L is this the expected structure for the `name` property on the entity?

CodeBaboon: I think it is `title` but I will try to confirm.

CodeBaboon: This is the activity-usage response for an html content topic on my local instance, appears to confirm `title` as the property name (assuming I'm doing this right):

image

test/activities/content/data/EditableContent.js
```diff
@@ -0,0 +1,10 @@
+export const editableContent = {
+	'class': [
+		'content'
```
CodeBaboon, line 3:
```suggestion
		'content-activity'
```

src/activities/content/ContentEntity.js
```diff
@@ -0,0 +1,13 @@
+import { Entity } from '../../es6/Entity';
+
+/**
+ * ContentEntity class representation of a d2l content entity.
+ */
+export class ContentEntity extends Entity {
```
CodeBaboon, line 6: Should these be `ContentEntity` or `ContentActivity`? Or even `ContentActivityEntity`?

test/activities/content/ContentEntity.js
```diff
@@ -0,0 +1,23 @@
...
+
+describe('ContentEntity', () => {
+	let editableEntity;
+
+	beforeEach(() => {
+		editableEntity = window.D2L.Hypermedia.Siren.Parse(editableContent);
+	});
+
+	afterEach(() => {
+		fetchMock.reset();
```
ianlaidlaw, line 14: what does `fetchMock` do in this file? do we need to reset it after every test?

imcdonald-d2l: It's not actually doing anything for us right now. It was a copy-paste from the ActivityEntity test file. Will remove it for now while we don't need it.

src/activities/content/ContentEntity.js
imcdonald-d2l: I think it will be personal preference on what are team thinks is best. Assignments is called `AssignmentEntity` so I'd be fine with either `ContentEntity` or `ContentActivityEntity` to stay in the same ballpark as assignments.

test/activities/content/data/EditableContent.js
imcdonald-d2l: Awesome, thanks for looking into that. Will adjust accordingly 👍

src/activities/content/ContentEntity.js
CodeBaboon: That makes sense, `ContentEntity` probably is fine then, and it's in the `activities/content` path anyway.

src/activities/content/ContentEntity.js
```diff
@@ -0,0 +1,13 @@
...
+	/**
+	 * @returns {string} Name of the content item
+	 */
+	name() {
```
CodeBaboon, line 10: Does this need to become `title` now also?

imcdonald-d2l: Yes, that would probably be a more logical function name to keep everything consistent. Will adjust accordingly.

prajapac (approved):

src/activities/content/ContentEntity.js
```diff
@@ -0,0 +1,13 @@
+import { Entity } from '../../es6/Entity';
+
+/**
+ * ContentEntity class representation of a d2l content entity.
+ */
+export class ContentEntity extends Entity {
+	/**
+	 * @returns {string} Name of the content item
```
CodeBaboon, line 8:
```suggestion
	 * @returns {string} Title of the content item
```

CodeBaboon (approved):

pbrown-d2l (approved):

PR #216, opened 2020-09-15
Tracking and displaying completion for organizations

tenbradley:
Adding properties and functions to the OrganizationEntity to contain this logic in the entity.

Added for [US119610](https://rally1.rallydev.com/#/detail/userstory/416180425096?fdp=true).

### DIFF
--- a/src/organizations/OrganizationEntity.js
+++ b/src/organizations/OrganizationEntity.js
@@ -1,11 +1,12 @@
+import { Rels, Classes, Actions } from '../hypermedia-constants';
+import { ActivityUsageEntity } from '../activities/ActivityUsageEntity.js';
 import { Entity } from '../es6/Entity.js';
 import { SimpleEntity } from '../es6/SimpleEntity.js';
-import { AlertsEntity } from './AlertsEntity.js';
-import { Rels, Classes, Actions } from '../hypermedia-constants';
+import { performSirenAction } from '../es6/SirenAction';
 import { NotificationCollectionEntity } from '../notifications/NotificationCollectionEntity';
 import { SequenceEntity } from '../sequences/SequenceEntity.js';
-import { ActivityUsageEntity } from '../activities/ActivityUsageEntity.js';
-import { performSirenAction } from '../es6/SirenAction';
+
+import { AlertsEntity } from './AlertsEntity.js';
 
 export const classes = {
 	active: 'active',
@@ -219,5 +220,21 @@ export class OrganizationEntity extends Entity {
 		}
 
 		return this._entity.getLinkByRel(rel).href;
+	}
+
+	getCompletionAction(isNotTracked) {
+		let actionName = 'do-not-track-completion';
+		if (isNotTracked) {
+			actionName = 'track-completion';
+		}
+		return this._entity.getActionByName(actionName);
+	}
+
+	getDisplayAction(isNotDisplayed) {
+		let actionName = 'do-not-diplay-progress';
+		if (isNotDisplayed) {
+			actionName = 'display-progress';
+		}
+		return this._entity.getActionByName(actionName);
 	}
 }
--- a/src/organizations/OrganizationEntity.js
+++ b/src/organizations/OrganizationEntity.js
@@ -43,18 +43,18 @@ export class OrganizationEntity extends Entity {
 
 	get isCompletionTracked() {
 		if (this._entity && this._entity.hasActionByName('track-completion')) {
-			return true;
-		} else if (this._entity && this._entity.hasActionByName('do-not-track-completion')) {
 			return false;
+		} else if (this._entity && this._entity.hasActionByName('do-not-track-completion')) {
+			return true;
 		}
 		return undefined;
 	}
 
 	get isProgressDisplayed() {
 		if (this._entity && this._entity.hasActionByName('display-progress')) {
-			return true;
-		} else if (this._entity && this._entity.hasActionByName('do-not-display-progress')) {
 			return false;
+		} else if (this._entity && this._entity.hasActionByName('do-not-display-progress')) {
+			return true;
 		}
 		return undefined;
 	}
--- a/src/organizations/OrganizationEntity.js
+++ b/src/organizations/OrganizationEntity.js
@@ -41,7 +41,7 @@ export class OrganizationEntity extends Entity {
 		return this._entity && this._entity.properties && this._entity.properties.endDate;
 	}
 
-	get isCompletionTracked(){
+	get isCompletionTracked() {
 		if (this._entity && this._entity.hasActionByName('track-completion')) {
 			return true;
 		} else if (this._entity && this._entity.hasActionByName('do-not-track-completion')) {
@@ -50,7 +50,7 @@ export class OrganizationEntity extends Entity {
 		return undefined;
 	}
 
-	get isProgressDisplayed(){
+	get isProgressDisplayed() {
 		if (this._entity && this._entity.hasActionByName('display-progress')) {
 			return true;
 		} else if (this._entity && this._entity.hasActionByName('do-not-display-progress')) {
@@ -236,7 +236,6 @@ export class OrganizationEntity extends Entity {
 		if (!this._entity || !this._entity.hasLinkByRel(rel)) {
 			return;
 		}
-
 		return this._entity.getLinkByRel(rel).href;
 	}
 
--- a/src/organizations/OrganizationEntity.js
+++ b/src/organizations/OrganizationEntity.js
@@ -125,7 +125,7 @@ export class OrganizationEntity extends Entity {
 
 	sequenceLink() {
 		return this._entity && this._entity.hasLinkByRel('https://api.brightspace.com/rels/sequence') &&
-		this._entity.getLinkByRel('https://api.brightspace.com/rels/sequence').href;
+			this._entity.getLinkByRel('https://api.brightspace.com/rels/sequence').href;
 	}
 
 	organizationHomepageUrl() {
@@ -239,19 +239,13 @@ export class OrganizationEntity extends Entity {
 		return this._entity.getLinkByRel(rel).href;
 	}
 
-	getCompletionAction(isTracked) {
-		let actionName = 'track-completion';
-		if (isTracked) {
-			actionName = 'do-not-track-completion';
-		}
+	getCompletionAction(nowTracked) {
+		const actionName = nowTracked ? 'track-completion' : 'do-not-track-completion';
 		return this._entity.getActionByName(actionName);
 	}
 
-	getDisplayAction(isDisplayed) {
-		let actionName = 'diplay-progress';
-		if (isDisplayed) {
-			actionName = 'do-not-display-progress';
-		}
+	getDisplayAction(nowDisplayed) {
+		const actionName = nowDisplayed ? 'display-progress' : 'do-not-display-progress';
 		return this._entity.getActionByName(actionName);
 	}
 }
--- a/src/organizations/OrganizationEntity.js
+++ b/src/organizations/OrganizationEntity.js
@@ -41,6 +41,24 @@ export class OrganizationEntity extends Entity {
 		return this._entity && this._entity.properties && this._entity.properties.endDate;
 	}
 
+	get isCompletionTracked(){
+		if (this._entity && this._entity.hasActionByName('track-completion')) {
+			return true;
+		} else if (this._entity && this._entity.hasActionByName('do-not-track-completion')) {
+			return false;
+		}
+		return undefined;
+	}
+
+	get isProgressDisplayed(){
+		if (this._entity && this._entity.hasActionByName('display-progress')) {
+			return true;
+		} else if (this._entity && this._entity.hasActionByName('do-not-display-progress')) {
+			return false;
+		}
+		return undefined;
+	}
+
 	isAfterEndDate() {
 		var nowDate = Date.now();
 		var endDate = Date.parse(this.endDate());
@@ -222,18 +240,18 @@ export class OrganizationEntity extends Entity {
 		return this._entity.getLinkByRel(rel).href;
 	}
 
-	getCompletionAction(isNotTracked) {
-		let actionName = 'do-not-track-completion';
-		if (isNotTracked) {
-			actionName = 'track-completion';
+	getCompletionAction(isTracked) {
+		let actionName = 'track-completion';
+		if (isTracked) {
+			actionName = 'do-not-track-completion';
 		}
 		return this._entity.getActionByName(actionName);
 	}
 
-	getDisplayAction(isNotDisplayed) {
-		let actionName = 'do-not-diplay-progress';
-		if (isNotDisplayed) {
-			actionName = 'display-progress';
+	getDisplayAction(isDisplayed) {
+		let actionName = 'diplay-progress';
+		if (isDisplayed) {
+			actionName = 'do-not-display-progress';
 		}
 		return this._entity.getActionByName(actionName);
 	}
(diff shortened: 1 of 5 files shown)


src/organizations/OrganizationEntity.js
```diff
@@ -1,11 +1,12 @@
...
 import { SimpleEntity } from '../es6/SimpleEntity.js';
-import { AlertsEntity } from './AlertsEntity.js';
-import { Rels, Classes, Actions } from '../hypermedia-constants';
+import { performSirenAction } from '../es6/SirenAction';
 import { NotificationCollectionEntity } from '../notifications/NotificationCollectionEntity';
 import { SequenceEntity } from '../sequences/SequenceEntity.js';
-import { ActivityUsageEntity } from '../activities/ActivityUsageEntity.js';
-import { performSirenAction } from '../es6/SirenAction';
+
+import { AlertsEntity } from './AlertsEntity.js';
```
AKobets, line 9: why this is reordered ?

src/organizations/OrganizationEntity.js
```diff
@@ -217,7 +236,16 @@ export class OrganizationEntity extends Entity {
 		if (!this._entity || !this._entity.hasLinkByRel(rel)) {
 			return;
 		}
-
 		return this._entity.getLinkByRel(rel).href;
 	}
+
+	getCompletionAction(nowTracked) {
```
AKobets, line 242: should we instead have methods: applyCompletionTracking(enabed) and applyDislayProgress(enabled) and perform action here instead of delegating it to the user of the Organization entity ?

zinabat: I suggest `updateCompletionTracking(enabled)`

test/organizations/OrganizationCompletionEntity.js
```diff
@@ -0,0 +1,76 @@
+import { OrganizationEntity } from '../../src/organizations/OrganizationEntity.js';
```
zinabat, line 1: Good job coming up with your test cases!

src/organizations/OrganizationEntity.js
tenbradley: I don't recall touching this.

test/organizations/OrganizationCompletionEntity.js
tenbradley: Testing is something I have done A LOT of in the past.

src/organizations/OrganizationEntity.js
tenbradley: I can move the entirety of the logic to the OrganizationEntity if desired. If so, I like the update option. 

Side note: if this is being done, I do not know how to test for the action being performed.

tenbradley: [aa0e3b0](https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/216/commits/aa0e3b0982d6400e6144b75673d7fcf954375202) is a compromise to separate the logic of getting the action from performing the action. These functions are underscored and tests are updated to reflect that, but the update versions are there to request the action and perform it.

src/organizations/OrganizationEntity.js
```diff
@@ -217,7 +236,26 @@ export class OrganizationEntity extends Entity {
...
+		const actionName = enabled ? 'track-completion' : 'do-not-track-completion';
+		return this._entity.getActionByName(actionName);
+	}
+
+	_getDisplayAction(enabled) {
+		const actionName = enabled ? 'display-progress' : 'do-not-display-progress';
+		return this._entity.getActionByName(actionName);
+	}
+
+	async updateCompletionTracking(enabled) {
```
zinabat, line 252: These don't need to be async - they already return a `Promise`.

src/organizations/OrganizationEntity.js
zinabat: For testing actions, you don't need to actually check the functionality of the update - we assume the API works as expected. Here's a set of tests that can be used as an example for testing an action: https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/test/organizations/OrganizationAvailabilityEntity.js#L84

tenbradley: Thanks. Added the testing for actions in [ba169f8](https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/216/commits/ba169f8b2fc43768a8d5d42bf27d0ec52b03644a).

src/organizations/OrganizationEntity.js
zinabat: The old one wasn't correct either. We generally alphabetize imports, with multi-imports appearing first. I added a commit to correct this.

zinabat (approved):

test/organizations/OrganizationCompletionEntity.js
```diff
@@ -0,0 +1,196 @@
...
```
AKobets, line 1: suggest renaming file to CompletionTracking.js
organization is already in the file path. And there is no class with name OrganizationCompletionEntity

test/organizations/data/OrganizationCompletionEntity.js
```diff
@@ -0,0 +1,233 @@
+export const testData = {
```
AKobets, line 1: Suggect renaming file to CompletionTracking.js, see explanation in similar comments above.

test/organizations/data/OrganizationCompletionEntity.js
```diff
@@ -0,0 +1,233 @@
...
+			{
+				'rel': [
+					'https://api.brightspace.com/rels/parent-semester'
+				],
+				'href': '../data/semester.json'
+			}
+		],
+		'actions': [
+			{
+				'href': '../completion/tracking.json',
```
AKobets, line 50: Does this file `../completion/tracking.json` need to exist?

tenbradley: No. I stole the info from the organizations repo. Since we never actually check the href, I figured this was fine. It can be changed if desired.

test/organizations/OrganizationCompletionEntity.js
```diff
@@ -0,0 +1,196 @@
...
+
+		const noTrackingJson = window.D2L.Hypermedia.Siren.Parse(testData.noTracking);
+		noTrackingDisplay = new OrganizationEntity(noTrackingJson);
+
+		const noActionsJson = window.D2L.Hypermedia.Siren.Parse(testData.noActions);
+		noActions = new OrganizationEntity(noActionsJson);
+	});
+	describe('properties working', () => {
+		it('tracking, no display', () => {
+			expect(trackingNoDisplay.isCompletionTracked).to.be.true;
```
AKobets: ideally unittest should assert behavior of one independend method.
If `trackingNoDisplay.isCompletionTracked` breaks, the test will not verify `trackingNoDisplay.isProgressDisplayed`

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,197 @@
...
+			it('incorrect parameters', () => {
+				display.updateCompletionTracking(true);
+				display.updateDisplayProgress(false);
+				expect(trackingSpy.returnValues[0]).to.be.undefined;
+				expect(displaySpy.returnValues[0]).to.be.undefined;
+			});
+		});
+	});
+
+	describe('update actions', () => {
```
AKobets, line 108: there is already test above with title `'update actions'` . Could you come up with unique names?

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,197 @@
...
+			expect(noActions._getDisplayAction(false)).to.be.undefined;
+		});
+	});
+
+	describe('update actions', () => {
+		let sandbox, trackingSpy, displaySpy, display;
+
+		beforeEach(() => {
+			sandbox = sinon.sandbox.create();
+			display = trackingNoDisplay;
```
AKobets, line 83: why it needs temporary variable display to hold reference to trackingNoDisplay

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,197 @@
...
+
+		afterEach(() => {
+			sandbox.restore();
+		});
+
+		describe('tracking completion, not tracking display', () => {
+			it('correct parameters', () => {
+				display.updateCompletionTracking(false);
+				display.updateDisplayProgress(true);
+				expect(trackingSpy.returnValues[0]).to.be.a('promise');
```
AKobets, line 96: it is also need  to assert that performAction was called with correct parameters.
I would suggest to split testing updateCompletionTracking and updateDisplayProgress into independent tests

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,197 @@
...
+			expect(noTrackingDisplay._getCompletionAction(true).name).to.equal('track-completion');
+			expect(noTrackingDisplay._getCompletionAction(false)).to.be.undefined;
+		});
+		it('no actions', () => {
+			expect(noActions._getCompletionAction(true)).to.be.undefined;
+			expect(noActions._getCompletionAction(false)).to.be.undefined;
+		});
+	});
+
+	describe('organization get display action', () => {
```
AKobets, line 59: _getDisplayAction is "private" function and is indirectly tested by updateDisplayProgress.
Maye be these test could be removed.

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,197 @@
...
+			expect(noTrackingDisplay.isProgressDisplayed).to.be.false;
+		});
+		it('actions missing', () => {
+			expect(noActions.isCompletionTracked).to.be.undefined;
+			expect(noActions.isProgressDisplayed).to.be.undefined;
+		});
+	});
+
+	describe('organization get completion action', () => {
+		it('tracking, no display', () => {
```
AKobets, line 41: _getCompletionAction is "private" function and is indirectly tested by updateCompletionTracking.
May be these tests could be removed.

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,197 @@
...
+
+		const noTrackingJson = window.D2L.Hypermedia.Siren.Parse(testData.noTracking);
+		noTrackingDisplay = new OrganizationEntity(noTrackingJson);
+
+		const noActionsJson = window.D2L.Hypermedia.Siren.Parse(testData.noActions);
+		noActions = new OrganizationEntity(noActionsJson);
+	});
+	describe('properties working', () => {
+		it('tracking, no display', () => {
+			expect(trackingNoDisplay.isCompletionTracked).to.be.true;
```
AKobets, line 23: I suggest testing isCompletionTracked and isProgressDisplayed independently from each other

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,197 @@
...
+			it('correct parameters', () => {
+				display.updateCompletionTracking(false);
+				display.updateDisplayProgress(true);
+				expect(trackingSpy.returnValues[0]).to.be.a('promise');
+				expect(displaySpy.returnValues[0]).to.be.a('promise');
+			});
+			it('incorrect parameters', () => {
+				display.updateCompletionTracking(true);
+				display.updateDisplayProgress(false);
+				expect(trackingSpy.returnValues[0]).to.be.undefined;
```
AKobets, line 102: also need to assert that performAction was not called.

tenbradley: How would one do that?

src/organizations/OrganizationEntity.js
```diff
@@ -217,7 +236,30 @@ export class OrganizationEntity extends Entity {
...
+	_getDisplayAction(enabled) {
+		const actionName = enabled ? 'display-progress' : 'do-not-display-progress';
+		return this._entity.getActionByName(actionName);
+	}
+
+	updateCompletionTracking(enabled) {
+		const action = this._getCompletionAction(enabled);
+		if (action) {
+			return performSirenAction(this._token, action, action.fields, false);
+		}
```
AKobets, line 256: if it returns promise in one case and undefined in another case you will not able to write `updateCompletionTracking().then()`
I think it should `return Promise.resolve(undefined);`

src/organizations/OrganizationEntity.js
```diff
@@ -40,6 +41,24 @@ export class OrganizationEntity extends Entity {
 		return this._entity && this._entity.properties && this._entity.properties.endDate;
 	}
 
+	get isCompletionTracked() {
+		if (this._entity && this._entity.hasActionByName('track-completion')) {
```
AKobets, line 45: there is actions object that defines names of actions. It is better to add actions for tracking completion and display progress there. you could reference them in new methods for example:`_entity.hasActionByName(actions.trackCompletion)`.

tenbradley:
@AKobets [ccab7bf](https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/216/commits/ccab7bff9d8a3d1a90b7fbeec9a785d596dd27c2) uses fetchMock to check that the href being called is the one expected or that no call has been made. I couldn't figure out how to check the other information for the called action but will look into it more.

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,278 @@
...
+			});
+
+			it('track completion', async() => {
+				const display = noTrackingDisplay;
+
+				await display.updateCompletionTracking(true);
+
+				expect(fetchMock.called()).to.be.true;
+				const call = await fetchMock.lastCall();
+				expect(call[0]).to.be.equal('http://api.x.io/track/completion');
```
AKobets, line 212: `expect(fetchMock.called('http://api.x.io/track/completion')).to.be.true;`  is more readable

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,278 @@
...
+				const form = await getFormData(call.request);
+				if (!form.notSupported) {
```
AKobets, line 214: If we remove form.notSupported check, will it pass testin in CI?
If we have this if statement, it is not clear if the assertion statement is ever executed

tenbradley: The form not supported check is because some browsers don't support it. I can try to remove it.

test/organizations/CompletionTracking.js
```diff
@@ -0,0 +1,278 @@
...
+		it('actions missing, completion', () => {
+			expect(noActions.isCompletionTracked).to.be.undefined;
+		});
+		it('actions missing, progress', () => {
+			expect(noActions.isProgressDisplayed).to.be.undefined;
+		});
+	});
+
+	describe('update actions, tracking completion, no display', () => {
+		let sandbox, trackingSpy, displaySpy, display;
```
AKobets, line 52: `display`  variable should be renamed to `organization`

AKobets (approved):

src/organizations/OrganizationEntity.js
```diff
@@ -40,6 +46,24 @@ export class OrganizationEntity extends Entity {
 		return this._entity && this._entity.properties && this._entity.properties.endDate;
 	}
 
+	get isCompletionTracked() {
```
m6jones, line 49: Although this makes sense as a `get` Does other entities do this now? If not I'd leave it as a method so people have the same experience.

src/organizations/OrganizationEntity.js
```diff
@@ -217,7 +240,32 @@ export class OrganizationEntity extends Entity {
...
+
+	updateCompletionTracking(enabled) {
+		const action = this._getCompletionAction(enabled);
+		if (action) {
+			return performSirenAction(this._token, action, action.fields, false);
+		}
+		return Promise.resolve(undefined);
+	}
+
+	updateDisplayProgress(enabled) {
```
m6jones, line 264: If you make this async you don't need to do `return Promise.resolve(undefined);`

PR #217, opened 2020-09-24
Don't call toString when scoreOutOf() returns undefined

imcdonald-d2l:
Issue in save workflow where `toString` is called on `scoreOutOf()` even when it returns undefined. This throws an error and the save workflow is not properly executed. Mitigating this by preventing `toString` being called when `scoreOutOf()` returns undefined.

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -635,9 +635,11 @@ export class ActivityUsageEntity extends Entity {
 			return;
 		}
 
+		const scoreOutOf = this.scoreOutOf() ? this.scoreOutOf().toString() : undefined;
+
 		return !this._shouldCreateAssociationToNewGrade(scoreAndGrade) &&
 			(this._shouldCreateAssociationToExistingGrade(scoreAndGrade)
-			|| scoreAndGrade.scoreOutOf !== this.scoreOutOf()?.toString()
+			|| scoreAndGrade.scoreOutOf !== scoreOutOf
 			|| scoreAndGrade.inGrades !== this.inGrades());
 	}
 
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -635,9 +635,11 @@ export class ActivityUsageEntity extends Entity {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -635,9 +635,11 @@ export class ActivityUsageEntity extends Entity {
 			return;
 		}
 
+		const scoreOutOf = this.scoreOutOf() ? this.scoreOutOf().toString() : undefined;
```
imcdonald-d2l, line 638: @wongvincent tagging you in this PR since you are the last to touch this code. Just confirming on your end that this is a non-issue for your team. I don't expect this to cause issues but want you to be aware. When saving a content activity-usage entity there is no related score information and `scoreOutOf()` will return `undefined` causing issues for us.

src/activities/ActivityUsageEntity.js
```diff
@@ -635,9 +635,11 @@ export class ActivityUsageEntity extends Entity {
...
+
 		return !this._shouldCreateAssociationToNewGrade(scoreAndGrade) &&
 			(this._shouldCreateAssociationToExistingGrade(scoreAndGrade)
-			|| scoreAndGrade.scoreOutOf !== this.scoreOutOf().toString()
+			|| scoreAndGrade.scoreOutOf !== scoreOutOf
```
CodeBaboon, line 642: Could the whole solution just be 
```suggestion
			|| scoreAndGrade.scoreOutOf !== this.scoreOutOf()?.toString()
```

ianlaidlaw, line 642: i haven't tested it out, but single not equals would work I think as it will just ignore one of them being a string, so conversion isn't necessary. but that might not be super clear as to why it does that kind of comparison.
```
scoreAndGrade.scoreOutOf != this.scoreOutOf()
```

ianlaidlaw: i wouldn't really recommend this though..

imcdonald-d2l: Yeah, not my code so don't want to mess with it too much. Don't want to blow up assignments stuff.

wongvincent (approved):

ianlaidlaw (approved):

PR #220, opened 2020-10-09
Rfegan/add quiz entity

ronan-f:
Part of [US119416](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/userstory/413005808740) new siren-sdk quiz entity to facilitate functionality related to updating quiz names.

### DIFF
--- a/src/activities/quizzes/QuizEntity.js
+++ b/src/activities/quizzes/QuizEntity.js
@@ -1,0 +1,36 @@
+import { Entity } from '../../es6/Entity';
+import { Actions } from '../../hypermedia-constants';
+import { performSirenAction } from '../../es6/SirenAction';
+
+/**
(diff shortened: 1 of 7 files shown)


mdgbayly (approved):
Looks good - when we get to doing the saving parts of this, we'll want to add a `save` implementation, similar to what we have in `AssignmentEntity` that allows us to execute the actions in bulk rather than saving each property one at a time. But that can come later.

ronan-f:
> Looks good - when we get to doing the saving parts of this, we'll want to add a `save` implementation, similar to what we have in `AssignmentEntity` that allows us to execute the actions in bulk rather than saving each property one at a time. But that can come later.

Yip that's what I was thinking. I intentionally tried to keep the scope of this PR as small as possible (i.e just the stuff that seemed relevant to editing a quiz name).

PR #221, opened 2020-10-19
Us119416/add quiz dirty check

ronan-f:
Setup for quizzes dirty check and part of [US119416](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=/userstory/413005808740).

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -183,7 +183,7 @@ export class ActivityUsageEntity extends Entity {
 	}
 
 	/**
- 	 * Updates the due date of the activity usage entity to the date specified
+	   * Updates the due date of the activity usage entity to the date specified
 	 * @param {string} dateValue Date string to set as the due date, or empty string to clear the due date
 	 */
 	async setDueDate(dateValue) {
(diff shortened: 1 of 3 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -664,10 +664,16 @@ export class ActivityUsageEntity extends Entity {
...
 			[this.endDate(), activity.dates.endDate],
 			[this.isDraft(), activity.isDraft],
-			[`${this.scoreOutOf()}`, activity.scoreAndGrade.scoreOutOf],
 			[this.inGrades(), activity.scoreAndGrade.inGrades]
 		];
 
+		const scoreOutOf = `${this.scoreOutOf()}`;
+
+		// For activities that don't use scoreOutOf, this.scoreOutOf() is initialized to undefined whereas for those that do it initializes to ''
+		if (scoreOutOf !== 'undefined') {
```
ronan-f, line 673: This feels quite hacky. The equals method here assumes every activity is using `scoreOutOf` which isn't necessarily the case. If an activity doesn't use `scoreOutOf` then the value of `this.scoreOutOf()` is set to 'undefined' which means the equals check will fail as `activity.scoreAndGrade.scoreOutOf` is an empty string even for activities without `scoreOutOf`

mdgbayly: Definitely kind of ugly - but i guess it was before too. Would it be better to just modify the `ActivityUsageEntity.scoreOutOf()` method to return an empty string if the scoreOutOf property is falsy?

mdgbayly (approved):
Seems fine but yeh the score out of handling is still kind of ugly as you mentioned. This is the same property that was causing issues with showing undefined in the page due to other changes to handle this property for tools that don't yet use score out of. I think there it was eventually changed to always initialize as a string, so might be better to do something similar here too.

PR #225, opened 2020-11-13
Renaming quiz subentity specific class to be more generic

wongvincent:
The reason for renaming the `hintsEnabled` constant, is that more quiz subentities are going to be added.

I also renamed the `enabled` string to `checked`. I am working on [US120983](https://rally1.rallydev.com/#/detail/userstory/438844241972?fdp=true) which adds the subentity "DisableRightClick". If DisableRightClick is checked, it can be confusing to include an `enabled` class, likewise a `disabled` class when it is unchecked (i.e. right click not actually disabled).
The backend change is in this PR: https://github.com/Brightspace/lms/pull/3446

Also, disabled could also make it confusing for the future, when someone thinks that class means the checkbox can't be editable or something.

### DIFF
--- a/src/activities/quizzes/QuizEntity.js
+++ b/src/activities/quizzes/QuizEntity.js
@@ -42,11 +42,11 @@ export class QuizEntity extends Entity {
 	}
 
 	/**
- * @returns {bool} Whether or not hints are enabled for the quiz entity
+ * @returns {bool} Whether or not hints are allowed for the quiz entity
  */
 	getHintsToolEnabled() {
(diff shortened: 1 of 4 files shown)


ronan-f (approved):
Looks good

mostlyfabulous:
@wongvincent Would this break existing stuff on legacy quizzing btw?

wongvincent:
Legacy quizzing isn't using this fancy hypermedia stuff, so it'll be fine!
yes, the `hintsEnabled` constant was the string `enabled`.

mostlyfabulous:
Right, forgot it isn't on hypermedia!

PR #226, opened 2020-11-16
Add content module delete action

imcdonald-d2l:
Adds delete action to `ContentModuleEntity` so we can call it from the activities UI side when the user cancels the creation of a new module.

**Related PRs**:
- https://github.com/BrightspaceHypermediaComponents/activities/pull/1184
- https://github.com/Brightspace/lms/pull/3414

create-cancel-delete

### DIFF
--- a/src/activities/content/ContentModuleEntity.js
+++ b/src/activities/content/ContentModuleEntity.js
@@ -88,9 +88,8 @@ export class ContentModuleEntity extends Entity {
 			return;
 		}
 
-		await performSirenAction(this._token, action).then(() => {
-			this.dispose();
-		});
+		await performSirenAction(this._token, action);
+		this.dispose();
 	}
 
 	/**
--- a/src/activities/content/ContentModuleEntity.js
+++ b/src/activities/content/ContentModuleEntity.js
@@ -88,7 +88,9 @@ export class ContentModuleEntity extends Entity {
 			return;
 		}
 
-		await performSirenAction(this._token, action, null);
+		await performSirenAction(this._token, action).then(() => {
+			this.dispose();
+		});
 	}
 
 	/**
--- a/src/activities/content/ContentModuleEntity.js
+++ b/src/activities/content/ContentModuleEntity.js
@@ -77,6 +77,21 @@ export class ContentModuleEntity extends Entity {
 	}
 
 	/**
+	 * Deletes the module
+	 */
+	async deleteModule() {
+		if (!this._entity) {
+			return;
+		}
+		const action = this._entity.getActionByName(Actions.content.deleteModule);
+		if (!action) {
+			return;
+		}
+
+		await performSirenAction(this._token, action, null);
+	}
+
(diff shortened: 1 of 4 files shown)


src/activities/content/ContentModuleEntity.js
```diff
@@ -76,6 +76,23 @@ export class ContentModuleEntity extends Entity {
...
+		if (!this._entity) {
+			return;
+		}
+		const action = this._entity.getActionByName(Actions.content.deleteModule);
+		if (!action) {
+			return;
+		}
+
+		await performSirenAction(this._token, action).then(() => {
+			this.dispose();
```
imcdonald-d2l, line 92: `this.dispose` goes [here](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/es6/Entity.js#L44). Similar pattern to what assignments is doing on delete. According to the function it deletes the callbacks and listeners stored in the entity store.

CodeBaboon (approved):

src/activities/content/ContentModuleEntity.js
ianlaidlaw: interesting to use a `.then` if we're already using an `asyc` function, imo.
> Code after each await expression can be thought of as existing in a .then callback. In this way a promise chain is progressively constructed with each reentrant step through the function. The return value forms the final link in the chain.

would doing 
```
await performSirenAction(this._token, action);
this.dispose();
```
not be the same?

ianlaidlaw (approved):

test/activities/content/ContentModuleEntity.js
```diff
@@ -81,5 +81,11 @@ describe('ContentModuleEntity', () => {
 			}
 			expect(fetchMock.called()).to.be.true;
 		});
+
+		it('deletes module', async() => {
```
ianlaidlaw, line 85: i might suggest to add another layer of descriptions for these tests.. and with better names. this doest really give a lot of info:
example:
```
describe('Actions', ...)
  describe('delete', ...)
    it('calls fetch mock or whatever you expect'
```

src/activities/content/ContentModuleEntity.js
imcdonald-d2l: That is correct, I believe that would do the same thing. I'll test that out.

test/activities/content/ContentModuleEntity.js
imcdonald-d2l: I don't personally know if that level of detail is really required here. I could change the test to `calls delete request` if this is an issue instead of adding another level of `describe` here. I know we aren't technically checking if a module is deleted but I think it gets the point across.

src/activities/content/ContentModuleEntity.js
imcdonald-d2l: Yeah, this works as expected and is cleaner. Have adjusted accordingly 👍

PR #238, opened 2020-12-02
US122707 Get preview href

ronan-f:
Part of https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F459179692392

### DIFF
--- a/test/activities/quizzes/QuizEntity.js
+++ b/test/activities/quizzes/QuizEntity.js
@@ -344,4 +344,18 @@ describe('QuizEntity', () => {
 			expect(fetchMock.done());
 		});
 	});
+
+	describe('previewHref', () => {
+		const testPreviewHref = 'http://test.desire2learn.d2l/d2l/lms/quizzing/user/quiz_summary.d2l?ou=6606&qi=46&isprv=1&fromQB=1&bp=1';
+
+		it('can read previewHref from an editable entity', () => {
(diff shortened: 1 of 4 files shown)


test/activities/quizzes/QuizEntity.js
```diff
@@ -344,4 +344,18 @@ describe('QuizEntity', () => {
...
+
+	describe('previewHref', () => {
+		const testPreviewHref = 'http://test.desire2learn.d2l/d2l/lms/quizzing/user/quiz_summary.d2l?ou=6606&qi=46&isprv=1&fromQB=1&bp=1';
+
+		it('can read previewHref from an editable entity', () => {
+			var quizEntity = new QuizEntity(editableEntity);
+			expect(quizEntity.previewHref()).to.equal(testPreviewHref);
+		});
+
+		it('can read previewHref from a non-editable entity', () => {
```
mostlyfabulous, line 356: The outcome for the `previewHref` isn't determined if an entity is editable or not but it probably is good to keep these 2 tests anyway?

mostlyfabulous (approved):
LGTM

test/activities/quizzes/QuizEntity.js
ronan-f: Yeah I mean the second one seems pretty redundant right now...I think it will be updated to something more meaningful once we add the permissions check

PR #243, opened 2020-12-04
Add PR reminder checklist

mostlyfabulous:
### Based off the Team USA PR checklist from the LMS repo

Feedback welcomed! Not sure if we should keep this as an opt-in and add some `if` logic to only target interested parties

**Current reminders:**

- Semver keywords in commit message

- Rally US URL if applicable

### DIFF
--- a/.github/workflows/prchecklist.yml
+++ b/.github/workflows/prchecklist.yml
@@ -1,11 +1,11 @@
-name: PR Checklist
+name: PR Reminder Checklist
 on:
   pull_request:
     types: [opened]
 jobs:
   fyi:
-    name: Posts a PR checklist
-    runs-on: [self-hosted, Lms, Linux]
+    name: Posts a PR Reminder checklist
+    runs-on: ubuntu-latest
     timeout-minutes: 2
     steps:
       - uses: Brightspace/third-party-actions@actions/github-script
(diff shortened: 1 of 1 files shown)


ronan-f:
I think this is a good addition as I keep forgetting the semantic release commit messages. I know a few teams maintain this repo so would like to hear input from others.

wongvincent:
I think this is a good addition as I keep forgetting the semantic release commit messages. I know a few teams maintain this repo so would like to hear input from others.

pureooze (approved):
:shipit:

m6jones (approved):
This check list specially the semver one is really cool! I am looking forward to this and may steal it for other repos.

pureooze:
Keep in mind there are also PR templates which you can use to do this kind of stuff if you want it to show up for all PRs. We used actions in the lms repo to avoid it being included for every single PR because it was a team specific checklist.

pureooze:
Thought about this more, might actually be best left as an action, that will allow it to show up on dependabot PRs for reviewers. I dont know how dependabot deals with templates 🤔

mostlyfabulous:
That's true, dependabot might break if there's a PR template conflicting with the PR description it generates.

Also, I wanted to keep this as an action so people can opt out of it if they want 😄

mostlyfabulous:
@ronan-f Oh the irony I forgot to use semver here lol.

PR #247, opened 2020-12-22
Add web-link definition to siren-sdk

imcdonald-d2l:
**NOTE:** I am on a new team starting on Jan 4th and would appreciate if someone could take over this PR for me. I'd be more than happy to help with clarification and add feedback where necessary but would be good if someone else could address the improvements/fixes necessary to get this merged.

Aligning siren-sdk with new work Dave Harris has added in the back-end [here](https://github.com/Brightspace/lms/pull/4949#event-4141743719). Adding all classes, properties and actions as defined so we can later access them in the activities repo.

Rally ticket: https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F463420383076

### DIFF
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -3,7 +3,8 @@ import { Rels } from '../../hypermedia-constants';
 
 export const CONTENT_TYPES = {
 	module: 'module',
-	topic: 'topic'
+	topic: 'topic',
+	weblink: 'weblink'
 };
 
 /**
@@ -12,17 +13,19 @@ export const CONTENT_TYPES = {
 export class ContentEntity extends Entity {
 
 	/**
-	 * @returns {string} content-module link
+	 * @returns {string} content type
 	 */
 	getEntityType() {
 		// TODO - add more specific entity types to the activity-usage class array
-		// once we start working on non-module content types
+		// as we start working on new content types
 		if (!this._entity) {
 			return null;
 		} else if (this._entity.hasClass(CONTENT_TYPES.module)) {
 			return CONTENT_TYPES.module;
 		} else if (this._entity.hasClass(CONTENT_TYPES.topic)) {
 			return CONTENT_TYPES.topic;
+		}  else if (this._entity.hasClass(CONTENT_TYPES.weblink)) {
+			return CONTENT_TYPES.weblink;
 		} else {
 			return null;
 		}
@@ -35,5 +38,14 @@ export class ContentEntity extends Entity {
 		return this._entity
 			&& this._entity.hasLinkByRel(Rels.Content.moduleEntity)
 			&& this._entity.getLinkByRel(Rels.Content.moduleEntity).href;
+	}
+
+	/**
+	 * @returns {string} content-weblink link
+	 */
+	getWebLinkHref() {
+		return this._entity
+			&& this._entity.hasLinkByRel(Rels.Content.weblinkEntity)
+			&& this._entity.getLinkByRel(Rels.Content.weblinkEntity).href;
 	}
 }
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -22,10 +22,10 @@ export class ContentEntity extends Entity {
 			return null;
 		} else if (this._entity.hasClass(CONTENT_TYPES.module)) {
 			return CONTENT_TYPES.module;
+		} else if (this._entity.hasClass(CONTENT_TYPES.weblink)) {
+			return CONTENT_TYPES.weblink;
 		} else if (this._entity.hasClass(CONTENT_TYPES.topic)) {
 			return CONTENT_TYPES.topic;
-		}  else if (this._entity.hasClass(CONTENT_TYPES.weblink)) {
-			return CONTENT_TYPES.weblink;
 		} else {
 			return null;
 		}
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -1,5 +1,6 @@
 import { Entity } from '../../es6/Entity';
 import { Rels } from '../../hypermedia-constants';
+import ContentHelperFunctions from './ContentHelperFunctions.js';
 
 export const CONTENT_TYPES = {
 	module: 'module',
@@ -20,9 +21,9 @@ export class ContentEntity extends Entity {
 		// as we start working on new content types
 		if (!this._entity) {
 			return null;
-		} else if (this._entity.hasClass(CONTENT_TYPES.module)) {
+		} else if (this._entity.hasLinkByRel(Rels.Content.moduleEntity)) {
 			return CONTENT_TYPES.module;
-		} else if (this._entity.hasClass(CONTENT_TYPES.weblink)) {
+		} else if (this._entity.hasLinkByRel(Rels.Content.weblinkEntity)) {
 			return CONTENT_TYPES.weblink;
 		} else if (this._entity.hasClass(CONTENT_TYPES.topic)) {
 			return CONTENT_TYPES.topic;
@@ -35,17 +36,13 @@ export class ContentEntity extends Entity {
 	 * @returns {string} content-module link
 	 */
 	getModuleHref() {
-		return this._entity
-			&& this._entity.hasLinkByRel(Rels.Content.moduleEntity)
-			&& this._entity.getLinkByRel(Rels.Content.moduleEntity).href;
+		return ContentHelperFunctions.getHrefFromRel(Rels.Content.moduleEntity, this._entity);
 	}
 
 	/**
 	 * @returns {string} content-weblink link
 	 */
 	getWebLinkHref() {
-		return this._entity
-			&& this._entity.hasLinkByRel(Rels.Content.weblinkEntity)
-			&& this._entity.getLinkByRel(Rels.Content.weblinkEntity).href;
+		return ContentHelperFunctions.getHrefFromRel(Rels.Content.weblinkEntity, this._entity);
 	}
 }
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -14,7 +14,7 @@ export const CONTENT_TYPES = {
 export class ContentEntity extends Entity {
 
 	/**
-	 * @returns {string} content type
+	 * @returns {string|null} content type
 	 */
 	getEntityType() {
 		// TODO - add more specific entity types to the activity-usage class array
--- a/src/activities/content/ContentModuleEntity.js
+++ b/src/activities/content/ContentModuleEntity.js
@@ -1,6 +1,7 @@
 import { Entity } from '../../es6/Entity';
-import { Actions, Classes } from '../../hypermedia-constants';
+import { Actions } from '../../hypermedia-constants';
 import { performSirenAction } from '../../es6/SirenAction';
+import ContentHelperFunctions from './ContentHelperFunctions.js';
 
 /**
  * ContentModuleEntity class representation of a d2l content-module entity.
@@ -11,28 +12,22 @@ export class ContentModuleEntity extends Entity {
 	 * @returns {string} Description html of the content-module item
 	 */
 	descriptionRichText() {
-		if (!this._entity) {
+		const descriptionSubEntity = ContentHelperFunctions.getDescriptionSubEntity(this._entity);
+		if (!descriptionSubEntity) {
 			return null;
 		}
-		const subEntity = this._entity.getSubEntitiesByClass(Classes.content.description)[0];
-		if (!subEntity || !subEntity.properties) {
-			return null;
-		}
-		return subEntity.properties.html || '';
+		return descriptionSubEntity.properties.html || '';
 	}
 
 	/**
 	 * @returns {string} Description text of the content-module item
 	 */
 	descriptionText() {
-		if (!this._entity) {
+		const descriptionSubEntity = ContentHelperFunctions.getDescriptionSubEntity(this._entity);
+		if (!descriptionSubEntity) {
 			return null;
 		}
-		const subEntity = this._entity.getSubEntitiesByClass(Classes.content.description)[0];
-		if (!subEntity || !subEntity.properties) {
-			return null;
-		}
-		return subEntity.properties.text || '';
+		return descriptionSubEntity.properties.text || '';
 	}
 
 	/**
@@ -83,7 +78,7 @@ export class ContentModuleEntity extends Entity {
 		if (!this._entity) {
 			return;
 		}
-		const action = this._entity.getActionByName(Actions.content.deleteModule);
+		const action = this._entity.getActionByName(Actions.module.deleteModule);
 		if (!action) {
 			return;
 		}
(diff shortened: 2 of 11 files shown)


test/activities/content/ContentWebLinkEntity.js
```diff
@@ -0,0 +1,148 @@
+/* global fetchMock */
+import { ContentWebLinkEntity } from '../../../src/activities/content/ContentWebLinkEntity.js';
+import { contentWebLinkData } from './data/TestContentWebLinkEntity.js';
+import { getFormData } from '../../utility/test-helpers.js';
+
+describe('ContentModuleEntity', () => {
```
ChrisKraljevicD2L, line 6:
```suggestion
describe('ContentWebLinkEntity', () => {
```

ChrisKraljevicD2L (commented):
This was the result of a group review with myself, @prajapac , @CodeBaboon and @brissons

src/activities/content/ContentEntity.js
```diff
@@ -12,15 +13,17 @@ export const CONTENT_TYPES = {
...
 	 */
 	getEntityType() {
 		// TODO - add more specific entity types to the activity-usage class array
-		// once we start working on non-module content types
+		// as we start working on new content types
 		if (!this._entity) {
 			return null;
 		} else if (this._entity.hasClass(CONTENT_TYPES.module)) {
 			return CONTENT_TYPES.module;
+		} else if (this._entity.hasClass(CONTENT_TYPES.weblink)) {
```
CodeBaboon, line 25: Just a note that currently the `activity-usage` is not returning a `weblink` class on it. I'm not sure if we want/need this or if the existence of the `content-weblink` Link is enough. @alamarre @ctwomey1 what you do you think? Should an `activity-usage` for a given activity have a class that specifies the type of the activity in a meaningful way?

CodeBaboon: This is currently the `activity-usage` response for a weblink activity:
image

I left the `topic` class on it because I don't want to break any existing consumers of these `activity-usage` entities. I did not specifically add `weblink` but I certainly can if that is desired.

imcdonald-d2l: Good to know! I agree, I can just replace it by checking for the existence of the `content-weblink` link.

src/activities/content/ContentEntity.js
```diff
@@ -36,4 +39,13 @@ export class ContentEntity extends Entity {
...
 	}
+
+	/**
+	 * @returns {string} content-weblink link
+	 */
+	getWebLinkHref() {
+		return this._entity
+			&& this._entity.hasLinkByRel(Rels.Content.weblinkEntity)
+			&& this._entity.getLinkByRel(Rels.Content.weblinkEntity).href;
+	}
```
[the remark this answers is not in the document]
imcdonald-d2l: So we get a `ContentEntity` when we get the activity usage and have to dig a layer deeper to get the actual information for either modules or weblinks by fetching the appropriate link. All the currently defined entities can be seen used in the `content.js` state file in activities [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-content-editor/state/content.js). This state file gets the `ContentEntity` [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-content-editor/state/content.js#L35) and then loads the relevant information depending on the type of content. If it is a module, it proceeds to load the `ContentModuleEntity` [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-content-editor/state/content.js#L44).

It is known that this state file will need to be further separated so that it doesn't get too complex with the addition of more entity types but that is the basic flow that I have currently setup. It didn't make sense to me to have `ContentEntity` as a superclass because all of the defined functions in `ContentEntity` would not work on `ContentModuleEntity` or `ContentWebLinkEntity`.

src/activities/content/ContentWebLinkEntity.js
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	/**
+	 * @returns {string} Description text of the content-weblink item
+	 */
+	descriptionText() {
+		if (!this._entity) {
+			return null;
+		}
+		const subEntity = this._entity.getSubEntitiesByClass(Classes.content.description)[0];
```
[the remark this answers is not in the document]
imcdonald-d2l: Interesting, did not know this syntax existed! Thanks 🙂

src/activities/content/ContentWebLinkEntity.js
```diff
@@ -0,0 +1,164 @@
...
+	 * @returns {string} Description html of the content-weblink item
+	 */
+	descriptionRichText() {
+		if (!this._entity) {
+			return null;
+		}
+		const subEntity = this._entity.getSubEntitiesByClass(Classes.content.description)[0];
+		if (!subEntity || !subEntity.properties) {
+			return null;
+		}
```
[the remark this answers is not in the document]
imcdonald-d2l: It is identical in `ContentModuleEntity` as well so I'll make a generic helper file with that function.

src/activities/content/ContentWebLinkEntity.js
```diff
@@ -0,0 +1,164 @@
...
+		return subEntity.properties.text || '';
+	}
+
+	/**
+	 * @returns {boolean} external resource value (i.e. open in new tab or not)
+	 */
+	isExternalResource() {
+		if (!this._entity) {
+			return null;
```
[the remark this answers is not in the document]
imcdonald-d2l: hmmm you bring up a good point. I'll change it to false. I doubt this condition will be hit when we call this function anyways but not sure if this would hide a potential error in the UI? I haven't setup any UI for the weblinks so can't confirm this yet.

src/activities/content/ContentWebLinkEntity.js
```diff
@@ -0,0 +1,164 @@
...
+	/**
+	 * @returns {string} Url of the content-weblink item
+	 */
+	url() {
+		return this._entity && this._entity.properties && this._entity.properties.url;
+	}
+
+	/**
+	 * Updates the web link to have the given description
+	 * @param {string} title rich text description to set on the web link
```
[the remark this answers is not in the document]
imcdonald-d2l: Good catch!

src/activities/content/ContentWebLinkEntity.js
```diff
@@ -0,0 +1,164 @@
...
+		if (!this._entity) {
+			return;
+		}
+		const action = this._entity.getActionByName(Actions.content.updateDescription);
+		if (!action) {
+			return;
+		}
+
+		const fields = [{ name: 'description', value: richText }];
+		await performSirenAction(this._token, action, fields);
```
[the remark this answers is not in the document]
imcdonald-d2l: It is set up this way because of the way the state is setup in the activities repo and how the various save functions are executed. When we save things we have to call the appropriate save actions from the various components in the editor (see modules save stuff [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-content-editor/state/content.js#L76)) and await their response so we know they execute successfully. These save functions are consolidated [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js#L145) and executed from this mixin [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js#L149). This mixin expects async functions and that is why all these functions that perform actions need to `await`.

test/activities/content/ContentWebLinkEntity.js
imcdonald-d2l: Good catch! 👀

src/activities/content/ContentEntity.js
ctwomey1: RE 
> Just a note that currently the activity-usage is not returning a weblink class on it. I'm not sure if we want/need this or if the existence of the content-weblink Link is enough. @alamarre @ctwomey1 what you do you think? Should an activity-usage for a given activity have a class that specifies the type of the activity in a meaningful way?

I think describing the entity fully would certainly make it more flexible for consumption.  It would all you to do things like pull all activities out of a collection with class "weblink" easier, but if current workflows don't require it, i wouldn't worry about it

ianlaidlaw:
I am taking over this pr - any questions can be addressed to me now. You won't even need to type a different name!

src/activities/content/ContentEntity.js
```diff
@@ -32,8 +36,13 @@ export class ContentEntity extends Entity {
...
-		return this._entity
-			&& this._entity.hasLinkByRel(Rels.Content.moduleEntity)
-			&& this._entity.getLinkByRel(Rels.Content.moduleEntity).href;
+		return ContentHelperFunctions.getHrefFromRel(Rels.Content.moduleEntity, this._entity);
+	}
+
+	/**
+	 * @returns {string} content-weblink link
+	 */
+	getWebLinkHref() {
```
ianlaidlaw, line 45: Maybe this way with a getter for every possibility is preferred, but I'm imagining that it's not possible to have existing hrefs for different entity types. Like, if it's a module entity it will have a module href and everything else will be null. so do we need a getter for everything or do we just need one that would check the `getModuleHref` and spit out what we expect? But like I said, maybe it's preferred to have everything compartmentalized like that... But it just feels like a bunch of redundant code that will ultimately return `null` or whatever in all but 1 scenario at each time

ianlaidlaw: I guess this does prevent us from attempting to access the href for a type that doesnt exist, but that can be gotten around by making the function take in an entity type and return the href based on that type. Is that  approach going against what this class is supposed to represent and would someone be able to explain to me why?

src/activities/content/ContentModuleEntity.js
```diff
@@ -11,28 +12,22 @@ export class ContentModuleEntity extends Entity {
...
 	 */
 	descriptionRichText() {
-		if (!this._entity) {
-			return null;
-		}
-		const subEntity = this._entity.getSubEntitiesByClass(Classes.content.description)[0];
-		if (!subEntity || !subEntity.properties) {
+		const descriptionSubEntity = ContentHelperFunctions.getDescriptionSubEntity(this._entity);
+		if (!descriptionSubEntity) {
 			return null;
```
ianlaidlaw, line 17: this function says `* @returns {string} Description html of the content-module item`
so either we need to return an empty string or update that ^

src/activities/content/ContentWebLinkEntity.js
```diff
@@ -0,0 +1,158 @@
...
+ */
+export class ContentWebLinkEntity extends Entity {
+
+	/**
+	 * @returns {string} Description html of the content-weblink item
+	 */
+	descriptionRichText() {
+		const descriptionSubEntity = ContentHelperFunctions.getDescriptionSubEntity(this._entity);
+		if (!descriptionSubEntity) {
+			return null;
```
ianlaidlaw, line 17: same thing here. need to check if empty string is okay, or if returning `null|string` is acceptable

ianlaidlaw: speaking from experience in some of our other repos, I really dislike the fallback of returning an empty string. `null` actually means something, the empty strings just hide problems.

ianlaidlaw: especially lines like this:
```
return descriptionSubEntity.properties.html || '';
```
I think having a safety net is fine, but I'd rather have it moved up and make us return `null` instead as that be able to quickly tell us there's problem with the api "why didn't descriptionSubEntity have properties, or if it did, why didn't it have an html if we expect it to exist?"

src/activities/content/ContentEntity.js
ianlaidlaw: I see @ctwomey1  kind of alluded to this above:
>I think describing the entity fully would certainly make it more flexible for consumption. It would all you to do things like pull all activities out of a collection with class "weblink" easier, but if current workflows don't require it, i wouldn't worry about it
do you think your comment applies to what I am describing as well?

CodeBaboon: This class (ContentEntity) represents the (somewhat) generic activity-usage object. When looking at the relevant siren response for a `module`, this is what we see:
image

The response for a `weblink` would be pretty similar, only the classes would change, and the link to the `content-module` entity would be replaced with a link to the `content-weblink` entity.

I think ideally we would have a different class for each type, though I'm not totally sure how that would be represented in this repo. I'm also not sure if it's really worth a bunch of refactoring to achieve it, considering the push to move to `hmc-components` going forward. Since that repo does not make use of siren-sdk any work we do in here to refactor and make this representation more perfect is potentially wasted effort that could be used to get us moved over to the `hmc-components` earlier.

src/activities/content/ContentEntity.js
CodeBaboon: It seems like it's maybe not immediately necessary but I agree that it probably would be better to include it. I'll look to add a `weblink` class as an independent PR sometime in the near future.

src/activities/content/ContentModuleEntity.js
imcdonald-d2l: Yeah, it's your call. I handle the null case in the UI I believe so returning an empty string isn't necessary and would hide errors like you mentioned somewhere else. So updating the comment would probably be preferred.

src/activities/content/ContentWebLinkEntity.js
imcdonald-d2l: Agreed, hiding errors isn't ideal so we can just update the comments and handle the errors at the UI level.

src/activities/content/ContentEntity.js
ianlaidlaw: Yeah I think it's fine the way it is now. Not that there can't be imporovements, but with the move to `hmc-foundation` It's not super important

src/activities/content/ContentEntity.js
```diff
@@ -12,15 +14,17 @@ export const CONTENT_TYPES = {
...
 	getEntityType() {
 		// TODO - add more specific entity types to the activity-usage class array
-		// once we start working on non-module content types
+		// as we start working on new content types
 		if (!this._entity) {
 			return null;
-		} else if (this._entity.hasClass(CONTENT_TYPES.module)) {
+		} else if (this._entity.hasLinkByRel(Rels.Content.moduleEntity)) {
 			return CONTENT_TYPES.module;
+		} else if (this._entity.hasLinkByRel(Rels.Content.weblinkEntity)) {
```
brissons, line 26: @CodeBaboon I'm pretty sure it's fine but wanted to double check, with [adding the weblink class](https://github.com/Brightspace/lms/pull/5172), this is the way to go?

CodeBaboon: I think this is fine. I'm not sure if we have something like `hasLinkByClass` or not in this SDK. If we do, we could switch to it, but the by rel is fine also. Basically adding the `weblink` class felt like the 'right' thing to do, as well as allows for some potentially better generic code down the line, but wasn't really intended to impact this code in the here-and-now.

CodeBaboon (approved):

ChrisKraljevicD2L (approved):

brissons (approved):

PR #249, opened 2021-01-05
support create-form link

lloydwaddell:
US122312

```Context```
Learning paths is adding a create-form link to redirect new activity usages to the different variation of the edit page. Siren-sdk activty entity needs to support the create-form rel.

https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F455165215100

```Quality```
Unit tests added to validate positive and negative response from createFormHref()

### DIFF
--- a/src/hypermedia-constants.js
+++ b/src/hypermedia-constants.js
@@ -204,7 +204,8 @@ export const Rels = {
 	},
 	IANA: {
 		edit: 'edit',
-		preview: 'preview'
+		preview: 'preview',
+		createForm: 'create-form'
 	}
 };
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/hypermedia-constants.js [hunk 1]
```diff
@@ -204,7 +204,8 @@ export const Rels = {
 	},
 	IANA: {
 		edit: 'edit',
-		preview: 'preview'
+		preview: 'preview',
```
AKobets, line 207: what `IANA` means ?

lloydwaddell: looks like its some sort of standard authority? 
https://www.iana.org/assignments/link-relations/link-relations.xhtml

AKobets (approved):

m6jones (approved):

PR #250, opened 2021-01-08
US123706 added AttemptsEntity and basic tests

me-louie:
Created a new AttemptsEntity class and wrote some basic tests

[US123706](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F478977336380&fdp=true?fdp=true): [ui] Attempts: Conditions entities

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F478977336380&fdp=true?fdp=true

### DIFF
--- a/src/activities/quizzes/AttemptsEntity.js
+++ b/src/activities/quizzes/AttemptsEntity.js
@@ -37,9 +37,7 @@ export class AttemptsEntity extends Entity {
 		if (!quiz) return;
 		if (!this._hasAttemptsAllowedChanged(quiz.attemptsAllowed)) return;
 
-		const attemptsAction = this._generateAttemptsAllowedAction(quiz.attemptsAllowed);
-
-		return attemptsAction;
+		return this._generateAttemptsAllowedAction(quiz.attemptsAllowed);
 	}
 
 	_hasAttemptsAllowedChanged(attemptsAllowed) {
--- a/src/activities/quizzes/AttemptsEntity.js
+++ b/src/activities/quizzes/AttemptsEntity.js
@@ -1,0 +1,87 @@
+import { Entity } from '../../es6/Entity';
+import { Actions } from '../../hypermedia-constants';
+import { performSirenAction } from '../../es6/SirenAction';
+
+/**
+ * AttemptsEntity class representation of a d2l Quiz Attempt.
+*/
+
+export class AttemptsEntity extends Entity {
+
+	/**
+	 * @returns {number} number of quiz attempts allowed
+	 */
+
+	attemptsAllowed() {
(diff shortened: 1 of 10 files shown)


src/activities/quizzes/AttemptsEntity.js
```diff
@@ -0,0 +1,87 @@
...
+	/**
+	 * Checks if quiz attempts has changed and if so returns the appropriate action/fields to update
+	 * @param {object} quiz the quiz that's being modified
+	 */
+
+	_formatUpdateAttemptsAction(quiz) {
+		if (!quiz) return;
+		if (!this._hasAttemptsAllowedChanged(quiz.attemptsAllowed)) return;
+
+		const attemptsAction = this._generateAttemptsAllowedAction(quiz.attemptsAllowed);
```
emil-furniss, line 40: nit: `return this._generateAttemptsAllowedAction(quiz.attemptsAllowed);`

src/activities/quizzes/QuizEntity.js
```diff
@@ -506,6 +507,10 @@ export class QuizEntity extends Entity {
 		});
 	}
 
+	attempts() {
```
emil-furniss, line 510: nit: Rename to `attemptsLink`

src/activities/quizzes/QuizEntity.js
```diff
@@ -480,7 +480,8 @@ export class QuizEntity extends Entity {
 			[this.password(), quiz.password],
 			[this.notificationEmail(), quiz.notificationEmail],
 			[this.isPreventMovingBackwardsEnabled(), quiz.preventMovingBackwards],
-			[this.isAutoSetGradedEnabled(), quiz.autoSetGraded]
+			[this.isAutoSetGradedEnabled(), quiz.autoSetGraded],
+			[JSON.stringify(this.attempts()), JSON.stringify(quiz.attempts)]
```
emil-furniss, line 484: I might be wrong here (maybe @wongvincent can chip in) but I don't think we to include attempts here. This equals function is primarily for comparing properties on the quiz entity for the dirty check, but the actual attempts entity is separate while the quiz only has a simple attempts link.

wongvincent: Yeah, since we'll have a separate siren-sdk entity for attempts and timing, we can have dirty checks in those entities instead.
Let's worry about that later.

me-louie: ok thanks, I will remove the attempts comparison from the equals function

emil-furniss (approved):

PR #251, opened 2021-01-11
Mdulat iwallace/us119424 UI quiz description field

mdulat:
https://rally1.rallydev.com/#/29180338367ud/iterationstatus?detail=%2Fuserstory%2F413008461552&fdp=true

### DIFF
--- a/src/activities/quizzes/QuizEntity.js
+++ b/src/activities/quizzes/QuizEntity.js
@@ -197,6 +197,78 @@ export class QuizEntity extends Entity {
 		return entity && entity.hasClass(Classes.quizzes.checked);
 	}
 
+	/**
+	 * @returns {string} Quiz discription in plaintext (HTML stripped)
+	 */
+	descriptionPlaintext() {
+		const descriptionEntity = this._getDescriptionEntity();
+		return descriptionEntity
+			&& descriptionEntity.properties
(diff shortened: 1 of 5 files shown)


src/activities/quizzes/QuizEntity.js
```diff
@@ -197,6 +197,78 @@ export class QuizEntity extends Entity {
...
+	 */
+	descriptionEditorHtml() {
+		const descriptionEntity = this._getDescriptionEntity();
+		if (!descriptionEntity) {
+			return;
+		}
+
+		const updateDescriptionAction = descriptionEntity.getActionByName(Actions.quizzes.updateDescription);
+		return updateDescriptionAction
+			&& updateDescriptionAction.hasFieldByName('description')
```
ronan-f, line 231: Do we want to make "description" a constant?

ronan-f (approved):
LGTM. Noticed that the string "description" is being used a few times so might be worth adding it as a constant?

src/activities/quizzes/QuizEntity.js
mdulat: That's a good point, maybe all the field names should be defined as constants? Makes sense for 'description' as it's used a few times, but it might be cleaner to be consistent with all field names. Maybe a clean-up task for a separate PR.

PR #265, opened 2021-02-04
Added support for IP restrictions

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F465115111280

Start of siren-sdk work for IP restrictions. This is pretty rough. Will need to add tests etc but wanted to get some feedback on the general approach.

### DIFF
--- a/src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
+++ b/src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
@@ -48,7 +48,7 @@ export class QuizIpRestrictionsEntity extends Entity {
 	 * @returns {object} Returns IP range sub entity
 	 */
 
-	getIpRestrictionSubEntity(id) {
+	_getIpRestriction(id) {
 		const ipRestrictionEntities = this._entity && this._entity.getSubEntitiesByClass(Classes.quizzes.ip.range);
 
 		if (!ipRestrictionEntities || !ipRestrictionEntities.length) {
@@ -93,7 +93,7 @@ export class QuizIpRestrictionsEntity extends Entity {
 	async deleteIpRestriction(id) {
 		if (!this.canEditIpRestrictions()) return;
 
-		const entity = this.getIpRestrictionSubEntity(id);
+		const entity = this._getIpRestriction(id);
 
 		if (!entity) return;
 
@@ -113,7 +113,7 @@ export class QuizIpRestrictionsEntity extends Entity {
 
 		const { start, end } = restriction;
 
-		const entity = this.getIpRestrictionSubEntity(restriction.id);
+		const entity = this._getIpRestriction(restriction.id);
 
 		if (!entity) {
 			return;
--- a/src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
+++ b/src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
@@ -33,15 +33,13 @@ export class QuizIpRestrictionsEntity extends Entity {
 			return [{ start: '', end: '' }];
 		}
 
-		return ipRestrictionEntities.map(({ properties }) => {
+		return ipRestrictionEntities.map(({ properties }, index) => {
 
-			const { start, end, oldStart, oldEnd } = properties;
-
+			const { start, end } = properties;
 			return {
 				start,
 				end,
-				oldStart,
-				oldEnd
+				id: index
 			};
 		});
 	}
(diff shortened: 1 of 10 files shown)


src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
```diff
@@ -0,0 +1,101 @@
...
+	 */
+
+	getIpRestrictions() {
+		if (!this._entity) {
+			return;
+		}
+
+		const ipRestrictionEntities = this._entity.getSubEntitiesByClass(Classes.quizzes.ip.range);
+
+		if (!ipRestrictionEntities || !ipRestrictionEntities.length) {
```
ronan-f, line 32: If there are no current IP restrictions then I believe we want the table to render a single empty row which the user can fill in. Not sure if this is the best place to make that decision.

src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
```diff
@@ -0,0 +1,101 @@
...
+
+		const { start, end, oldStart, oldEnd } = restriction;
+
+		const action = this._entity.getActionByName(Actions.quizzes.ipRestrictions.add);
+		const fields = [
+			{ name: IP_START_FIELD, value: start },
+			{ name: IP_END_FIELD, value: end }
+		];
+
+		if (oldStart && oldEnd) {
```
ronan-f, line 74: This is where we decide to update or add a new restriction. If the restriction has oldStart and oldEnd values then the API will interpret it as an update. These values won't be present for new restrictions

src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
```diff
@@ -0,0 +1,101 @@
...
+
+		const ipRestrictionEntities = this._entity.getSubEntitiesByClass(Classes.quizzes.ip.range);
+
+		if (!ipRestrictionEntities || !ipRestrictionEntities.length) {
+			return [{ start: '', end: '' }];
+		}
+
+		return ipRestrictionEntities.map(({ properties }) => {
+
+			const { start, end, oldStart, oldEnd } = properties;
```
mdgbayly, line 38: This observation is going to affect how you handle this data in the MobX state, and I have some other comments coming about that, but I wasn't intending that we include the `old` values as properties. These are effectively hidden hypermedia state in the field definition that is put there just so that the server knows how to process the action. The client shouldn't be aware it even exists, which is why the `old` fields should be defined as `type="hidden"` in the action. The siren action implementation will automatically include those hidden fields when the client executes that action without the client having to deal with them at all.

This does have other implications for how you are managing the state and how you associate what you are managing in the MobX state with what is represented in the original entity, as you need some way to related the MobX state back to the relevant entity in the original siren response. 

I hadn't fully thought this out, but just thinking about it now, I was wondering if we could use the index of each existing IP restriction in the siren entity as a kind of identifier, that the MobX uses to track which bit of state it is showing in the UI is associated with which entity in the Siren response. This wouldn't be used as an offset based index in the MobX state, because it would be possible to delete an entry in the UI such that what is at position 2 in the UI might be associated with a siren entity at position 1 for example.

e.g. when you return the list of IP restrictions to the MobX state you would return something that looked more like this:

[
{ "start": "192.168.1.1", "end": "192.168.1.10", "id": 0},
{ "start": "192.168.1.20", "end": "192.168.1.30", "id": 10}
]

etc.

New IP restrictions that get added wouldn't have an id in the MobX state which is how you can track that they are new.

I think this would affect the implementation of add and delete below, and we'd probably also want an update that utilized the id field as a property to the update method, so it could find the relevant sub entity to update. The sub entity update action would include the `old` hidden fields, and so no need to be passing them in the `update` method call.

src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
mdgbayly: As discussed above, shouldn't need to pass `old` values, if we track the associated IP restriction sub-entity using an identifier field instead.

src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	/**
+	 * @param {string} start Start of IP range to delete
+	 * @param {string} end End of IP range to delete
+	 * @returns {null}
+	*/
+
+	async deleteIpRestriction(start, end) {
+		if (!this.canEditIpRestrictions()) return;
```
mdgbayly, line 91: Did we discuss moving the delete action to the individual IP restriction entities rather than having a generic delete on the collection. Wouldn't this be needed anyway, in case someone modifies an existing IP restriction before deleting it? The current start/end in the UI might not equate to the actual back-end values. So if we had the `id` property as discussed in the other comments we could delete just by id, which the siren-sdk would use the find the appropriate sub-entity by offset, and call it's sub entity delete action which would include the original start/end values.
I guess that also means that the start/end values in those delete actions should probably be `type="hidden"` too.

src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
```diff
@@ -0,0 +1,130 @@
...
+
+	async deleteIpRestriction(id) {
+		if (!this.canEditIpRestrictions()) return;
+
+		const entity = this._getIpRestriction(id);
+
+		if (!entity) return;
+
+		const action = entity.getActionByName(Actions.quizzes.ipRestrictions.delete);
+		const fields = [
```
mdgbayly, line 101: You shouldn't need to copy these fields from the properties? Doesn't it just keep the fields that were already in the action? Maybe you just need to pass null/undefined for the fields?

mdgbayly (approved):
Looks good - just that one small nit on the Delete implementation. Can be fixed on a follow up PR.

PR #270, opened 2021-02-09
Cleaned up IP restrictions delete

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F465115111280

We actually don't need to specify these fields as they are provided by the action.


me-louie (approved):
LGTM

PR #273, opened 2021-02-19
US123360 - Update how automatic-zero is detected

wongvincent:
With optimistic rendering, both enforced and recommended entities and this "uselatelimit/automaticzero" entity is always included. So we must determine whether to show it based on whether the enforced timing entity type has the class, rather than the presence of the sub-entity.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F464625055264%2Ftasks


ronan-f (approved):
Makes sense to me

PR #280, opened 2021-03-02
US123692 - Fix EntityStore updating incorrect entity

wongvincent:
A base quiz entity contains a checkout action with a quiz href. After POSTing to it, the response is the working copy entity that ends with a workingCopy queryParam. Wtihout my change, the EntityStore was updating the base quiz entity with the returned working copy entity, when in fact, it should be adding a new object to the entity store with the self href of the returned working copy entity.

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F478606416460%2Ftasks

### DIFF
--- a/src/es6/SirenAction.js
+++ b/src/es6/SirenAction.js
@@ -146,6 +146,10 @@ const _performSirenAction = function(action, fields, tokenValue) {
 				if (!entity) {
 					return window.D2L.Siren.EntityStore.remove(url.href, token);
 				}
+				const selfLink = entity.getLinkByRel('self');
+				if (selfLink && selfLink.href !== url.href) {
+					return window.D2L.Siren.EntityStore.update(selfLink.href, token, entity);
+				}
 				return window.D2L.Siren.EntityStore.update(url.href, token, entity);
 			});
 		});


wongvincent:
This problem was found when we loaded an existing quiz, then clicked 'save'. All the existing quiz input fields become disabled. This happens because clicking 'save' refetches the quiz from the EntityStore, which has the wrong entity as the working copy entity ended up being stored in the (base) quiz spot.

src/es6/SirenAction.js
```diff
@@ -146,6 +146,10 @@ const _performSirenAction = function(action, fields, tokenValue) {
 				if (!entity) {
 					return window.D2L.Siren.EntityStore.remove(url.href, token);
 				}
+				const selfLink = entity.getLinkByRel('self');
+				if (selfLink && selfLink.href !== url.href) {
+					return window.D2L.Siren.EntityStore.update(selfLink.href, token, entity);
+				}
 				return window.D2L.Siren.EntityStore.update(url.href, token, entity);
```
wongvincent, line 153: To maintain the existing behavior, it's possible someone is using this and returning an entity missing a `self` href, so I will update the store with the returned entity and original request href if the self href is missing.

src/es6/SirenAction.js
```diff
@@ -146,6 +146,10 @@ const _performSirenAction = function(action, fields, tokenValue) {
 				if (!entity) {
 					return window.D2L.Siren.EntityStore.remove(url.href, token);
 				}
+				const selfLink = entity.getLinkByRel('self');
```
mdgbayly, line 149: hmm, a little bit worrying that this wasn't already standard behaviour...

I guess it's highly unlikely? that someone would return an href with a different self link and still want it to update the store with the request url, but I'm a bit concerned that someone might have a wacky use case like that.

Probably worthwhile to advertise this change on #web-platform just in case, but I feel like it makes sense given how things like the Siren hypermedia browser works.

minor nit: could avoid the duplicate line `window.D2L.Siren.EntityStore.update` if you stored the href in a local variable.

mdgbayly (approved):
LGTM but recommend advertise this change.

PR #285, opened 2021-03-04
Add support for working copy actions to web link siren entities

CodeBaboon:
## Description
​
Working copy actions (checkout/commit/cancel) have been added to the weblinks domain siren entities via https://github.com/Brightspace/lms/pull/6280. In order for the Activities user interface to interact with these new actions we need to modify the siren-sdk entity representation accordingly.
​
## Goal/Solution
​
This PR simply adds the new working copy actions to the WebLink entity. It also modifies some of the other actions to more accurately return their async responses.

## Video/Screenshots
​
N/A

​
## Remaining Work
- [x] Remove `cancel` action from weblink entity (see [here](https://github.com/Brightspace/lms/pull/6280/commits/4d02f64f7b7410d1444a752744b5e1b70b10c884) for details )
​​

## Relevant Rally Links 
​
[US124553](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F490369835660&fdp=true?fdp=true): Weblinks > Use working copy during FACE create workflow


src/activities/content/ContentWebLinkEntity.js
```diff
@@ -68,7 +68,7 @@ export class ContentWebLinkEntity extends Entity {
 		}
 
 		const fields = [{ name: 'description', value: richText }];
-		await performSirenAction(this._token, action, fields);
+		return await performSirenAction(this._token, action, fields);
```
ianlaidlaw, line 71: good change.. async functions are required to return a Promise iirc. i dont think these functions were really doing the async behavior that it _appeared_ like they were doing.

src/hypermedia-constants.js
```diff
@@ -508,6 +508,9 @@ export const Actions = {
 		updateExternalResource: 'update-external-resource',
 		deleteWeblink: 'delete-webLink',
 		deleteLTIlink: 'delete-ltiLink',
+		checkoutWebLink: 'checkout',
```
ianlaidlaw, line 511: just gonna document this here now, definitely no need to change or even worry about it yet for HTML documents.. but this looks like a good opportunity to make this pretty generic and have some sort of `"ContentWorkingCopyEntity"` that will just automatically get these actions and we won't need to rewrite and add them for every new type

ianlaidlaw (approved):

src/activities/content/ContentWebLinkEntity.js
```diff
@@ -138,6 +138,51 @@ export class ContentWebLinkEntity extends Entity {
...
+			return;
+		}
+
+		return await performSirenAction(this._token, action);
+	}
+
+	/**
+	 * performs a cancel action on the web link
+	 */
+	async cancelWebLink() {
```
ianlaidlaw, line 159: Maybe I'm being overly explicit here, but IMO maybe these function names should have `working copy` somewhere in there. I was reading the `activities` pr and it can be _slightly_ confusing that the `cancel` refers to working copy specifically.

src/hypermedia-constants.js
CodeBaboon: Yeah this is a good call, and something I believe @quinnbudan suggested also. Seems like a good refactor for a future PR .

src/activities/content/ContentWebLinkEntity.js
CodeBaboon: I see what you're saying, that's probably a good call. Should we do that in this PR (and the associated Activities PR) or as a follow-up?

ianlaidlaw: Followup seems fine, no reason to hold this up

brissons (approved):
lgtm!

prajapac (approved):

src/activities/content/ContentWebLinkEntity.js
CodeBaboon: @ianlaidlaw , now that the `cancel` action no longer exists, do you still feel as strongly about adding `workingCopy` to the others? I feel like `checkout` and `commit` are a lot more explicit than cancel was and it is maybe less necessary now, but interested in your thoughts.

ianlaidlaw: seems fine to me!

PR #286, opened 2021-03-04
Default IPs to empty strings to prevent weird errors

ronan-f:
If IPs were ever loaded from LMS as `undefined` weird things happened...


mostlyfabulous (approved):

mostlyfabulous:
When does that behaviour happen btw?

ronan-f:
> When does that behaviour happen btw?

If the user only provides a start value, previously we set end = start but now we copy legacy and only save start if that's all that's provided. This means that siren tries to access `properties.end` which doesn't exist so it implicitly sets it to `undefined`

PR #289, opened 2021-03-05
Check if restriction has already been persisted

ronan-f:
This check is needed as without it, if a user changes the order of restrictions in local state, it won't realize that they are existing restrictions and will try to save them again in which case the server will flag them as duplicates. An alternative would be to remove that check server side but then we still fire off a bunch of unnecessary network requests. This uses the existing siren-sdk state to as a source of truth for what has already been persisted. If an IP from MobX is present in siren's state, regardless of the order, it should not be sent again.

### DIFF
--- a/src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
+++ b/src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
@@ -35,10 +35,10 @@ export class QuizIpRestrictionsEntity extends Entity {
 
 		return ipRestrictionEntities.map(({ properties }, index) => {
 
-			const { start = '', end = '' } = properties;
+			const { start, end } = properties;
 			return {
-				start,
-				end,
+				start: start || '',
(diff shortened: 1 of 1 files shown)


src/activities/quizzes/ipRestrictions/QuizIpRestrictionsEntity.js
```diff
@@ -35,10 +35,10 @@ export class QuizIpRestrictionsEntity extends Entity {
 
 		return ipRestrictionEntities.map(({ properties }, index) => {
 
-			const { start = '', end = '' } = properties;
+			const { start, end } = properties;
 			return {
-				start,
-				end,
+				start: start || '',
```
ronan-f, line 40: Changed this fall back to here as the default values before only checked for `undefined` which means an IP is sometimes `null`. This way should guarantee any null or undefined ranges will be treated as an empty string.

wongvincent (approved):

PR #290, opened 2021-03-08
Mlouie/us125529

me-louie:
As discussed with Martin, we modify  `siren-sdk` to have an internal field called `_linkedScoreOutOfEntity` . When the  linked `score-out-of` entity is present, we conditionally load it and use it. This way, the Quiz UI can use the `score-out-of value` in a manner that is consistent with Assignments (i.e. via the `ActivityScoreGrade`).

[US125529](https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F505601651720&fdp=true?fdp=true): [ui] Display total points for a quiz

https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F505601651720&fdp=true?fdp=true


me-louie:
made a new branch for the refactoring

PR #297, opened 2021-03-16
DE42885: Add more null checking on the scoreoutof entity

ianlaidlaw:
On the Content FACE pages, we use the generic `activity-usage`, however our entities have no concept of grades or scoreoutof. As a result, our FACE pages were erroring out and dates were not working.  

This pr adds more aggressive null checking so we are not trying to get properties off of something that does not exist.

image

https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F511812392384&fdp=true


me-louie (approved):
Thanks for adding this

brissons (approved):

quinnbudan (approved):

PR #303, opened 2021-04-01
add functions to enable checkout of activity usage

mdulat:
https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fuserstory%2F505562070580&fdp=true

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -735,4 +735,52 @@ export class ActivityUsageEntity extends Entity {
 
 		return this._entity.getLinkByRel(Rels.Alignments.alignmentsHierarchical).href;
 	}
+
+	/**
+	* @returns {string} URL of the associate-grade API, for managing grade association with the activity usage, using working copy, if present
+	*/
+	associateGradeHref() {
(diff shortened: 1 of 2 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -735,4 +735,52 @@ export class ActivityUsageEntity extends Entity {
...
+	*/
+	associateGradeHref() {
+		if (!this._entity || !this._entity.hasLinkByRel(Rels.Activities.associateGrade)) {
+			return;
+		}
+
+		return this._entity.getLinkByRel(Rels.Activities.associateGrade).href;
+	}
+
+	_canCheckout() {
```
emil-furniss, line 750: Is the underscore intended while the `canCheckin` doesn't have one?

emil-furniss (approved):

src/activities/ActivityUsageEntity.js
mdulat: Yes, it's because `_canCheckout` is only used within this file, whereas `canCheckin` will be used directly by activities.

wongvincent (approved):

PR #304, opened 2021-04-05
US126237: Siren SDK updates for html files

brissons:
This PR adds the html file definition. Lots taken from @ianlaidlaw's and @CodeBaboon's branches. Ty!

`ContentHtmlFileEntity.js` and `FileLocationEntity.js` are the main definitions added, where `ContentHtmlFileEntity.js` has a lot of the same information, and is set up similar, to the other entities. `FileLocationEntity.js` contains the link which points to the actual html file data. 

It's set up right now that the activities repo will be in charge of populating these entities. For example, when `ContentHtmlFileEntity` is initialized, it has a` getFilesHref()` which, when the link is called, we will need to populate the `FileLocationEntity.js`. Once `FileLocationEntity.js` is initialized, we `getFileLocationHref()` to fetch the actual file data.

~~This PR is currently not genericized for files/is specific to html files. Further commits will be done in this PR for genericizing~~ ~~to all content files (or opened in a new PR if this one gets too messy) once it is done in the BE.~~

Edit: I will merge this one for html files, then make a separate one for genericizing. 

- [x] Finish tests

[Here is the rally ticket](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fuserstory%2F514758865880&fdp=true)

### DIFF
--- a/src/activities/content/ContentHtmlFileEntity.js
+++ b/src/activities/content/ContentHtmlFileEntity.js
@@ -1,0 +1,133 @@
+import { Actions, Rels } from '../../hypermedia-constants.js';
+import { performSirenAction } from '../../es6/SirenAction.js';
+import ContentHelperFunctions from './ContentHelperFunctions.js';
+import { ContentWorkingCopyEntity } from './ContentWorkingCopyEntity.js';
+
+/**
+ * ContentHtmlFileEntity class representation of a d2l content-htmlFile entity.
+ */
+export class ContentHtmlFileEntity extends ContentWorkingCopyEntity {
+
+	/**
+	 * @returns {string|null} Description html of the content-htmlFile item
+	 */
+	descriptionRichText() {
+		const descriptionSubEntity = ContentHelperFunctions.getDescriptionSubEntity(this._entity);
+		if (!descriptionSubEntity) {
+			return null;
+		}
+		return descriptionSubEntity.properties.html || '';
+	}
+
+	/**
+	 * @returns {string|null} Description text of the content-htmlFile item
+	 */
+	descriptionText() {
+		const descriptionSubEntity = ContentHelperFunctions.getDescriptionSubEntity(this._entity);
+		if (!descriptionSubEntity) {
+			return null;
+		}
+		return descriptionSubEntity.properties.text || '';
+	}
+
+	/**
+	 * @returns {string} Title of the content-htmlFile item
+	 */
+	title() {
+		return this._entity && this._entity.properties && this._entity.properties.title;
+	}
+
+	/**
+	 * Updates the html file to have the given description
+	 * @param {string} richText rich text description to set on the html file
+	 */
+	async setHtmlFileDescription(richText) {
+		if (!this._entity) {
+			return;
+		}
+		const action = this._entity.getActionByName(Actions.content.updateDescription);
+		if (!action) {
+			return;
+		}
+
+		const fields = [{ name: 'description', value: richText }];
+		await performSirenAction(this._token, action, fields);
+	}
+
+	/**
+	 * Updates the html file to have the given title
+	 * @param {string} title Title to set on the html file
+	 */
+	async setHtmlFileTitle(title) {
+		if (!this._entity) {
+			return;
+		}
+		const action = this._entity.getActionByName(Actions.content.updateTitle);
+		if (!action) {
+			return;
+		}
+
+		const fields = [{ name: 'title', value: title }];
+		await performSirenAction(this._token, action, fields);
+	}
+
+	/**
+	 * Updates the html file content with the given html
+	 * @param {html} html to set on the html file
+	 */
+	async setHtmlFileHtml(html) {
+		if (!this._entity) {
+			return;
+		}
+		const action = this._entity.getActionByName(Actions.htmlFile.updateHtmlContent);
+		if (!action) {
+			return;
+		}
+
+		const fields = [{ name: 'htmlContent', value: html }];
+		await performSirenAction(this._token, action, fields);
+	}
+
+	/**
+	 * Deletes the html file
+	 */
+	async deleteHtmlFile() {
+		if (!this._entity) {
+			return;
+		}
+		const action = this._entity.getActionByName(Actions.htmlFile.deleteHtmlFile);
+		if (!action) {
+			return;
+		}
+
+		await performSirenAction(this._token, action);
+		this.dispose();
+	}
+
+	/**
+	 * Checks if content html file properties passed in match what is currently stored
+	 * @param {object} contentHtmlFile Object containing html file specific properties
+	 */
+	equals(contentHtmlFile) {
+		const diffs = [
+			[this.title(), contentHtmlFile.title],
+		];
+		for (const [left, right] of diffs) {
+			if (left !== right) {
+				return false;
+			}
+		}
+		return true;
+	}
+
+	/**
+	* @returns {string} Returns URL of the files home API
+	*/
+	getFilesHref() {
+		if (!this._entity || !this._entity.hasLinkByRel(Rels.Files.file)) {
+			return;
+		}
+
+		return this._entity.getLinkByRel(Rels.Files.file).href;
+	}
+}
--- a/src/activities/content/ContentHtmlFileEntity.js
+++ b/src/activities/content/ContentHtmlFileEntity.js
@@ -75,7 +75,7 @@ export class ContentHtmlFileEntity extends ContentWorkingCopyEntity {
 	 * Updates the html file content with the given html
 	 * @param {html} html to set on the html file
 	 */
-	async setHtmlFileContent(html) {
+	async setHtmlFileHtmlContent(html) {
 		if (!this._entity) {
 			return;
 		}
--- a/src/activities/content/ContentHtmlFileEntity.js
+++ b/src/activities/content/ContentHtmlFileEntity.js
@@ -31,7 +31,7 @@ export class ContentHtmlFileEntity extends ContentWorkingCopyEntity {
 	}
 
 	/**
-	 * @returns {string} Title of the content-htmlFile item
+	 * @returns {string|undefined} Title of the content-htmlFile item
 	 */
 	title() {
 		return this._entity && this._entity.properties && this._entity.properties.title;
@@ -75,7 +75,7 @@ export class ContentHtmlFileEntity extends ContentWorkingCopyEntity {
 	 * Updates the html file content with the given html
 	 * @param {html} html to set on the html file
 	 */
-	async setHtmlFileHtml(html) {
+	async setHtmlFileContent(html) {
 		if (!this._entity) {
 			return;
 		}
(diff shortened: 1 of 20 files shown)


test/activities/content/ContentHtmlFileEntity.js
```diff
@@ -0,0 +1,104 @@
...
+
+		it('Equality should return false when title is different', () => {
+			const htmlFileData = {
+				title: 'New Title',
+				descriptionRichText: '<p>description text</p>'
+			};
+			expect(contentHtmlFileEntity.equals(htmlFileData)).to.equal(false);
+		});
+
+		it('Equality should return true when descriptions match', () => {
```
brissons, line 50: I left this in here to discuss something I noticed. We have descriptions (plain and rich) for each of our activities ("text" and "html"), although to my knowledge we don't use them. Right now, it looks like these descriptions are not included in the `equals()` methods, but we test for them (like this one). Ex: for LTI links, [we have the equals method](https://search.d2l.dev/xref/BrightspaceHypermediaComponents/siren-sdk/src/activities/content/ContentLTILinkEntity.js?r=7bc5a317#128) that does not include the description, but [we test for it here](https://search.d2l.dev/xref/BrightspaceHypermediaComponents/siren-sdk/test/activities/content/ContentLTILinkEntity.js?r=c62d75d3#62). I'm inclined to just delete these tests since we aren't using the descriptions, but I'm wondering what y'all think.

src/hypermedia-constants.js
```diff
@@ -110,7 +110,8 @@ export const Rels = {
 	Content: {
 		moduleEntity: 'https://modules.api.brightspace.com/rels/content-module',
 		weblinkEntity: 'https://weblinks.api.brightspace.com/rels/content-weblink',
-		ltilinkEntity: 'https://weblinks.api.brightspace.com/rels/content-ltilink'
+		ltilinkEntity: 'https://weblinks.api.brightspace.com/rels/content-ltilink',
+		htmlFileEntity: 'https://content.api.brightspace.com/rels/content-html-file'
```
brissons, line 114: This will likely change during genericization.

src/activities/content/ContentWorkingCopyEntity.js
```diff
@@ -0,0 +1,51 @@
+import { Entity } from '../../es6/Entity';
+import { Actions } from '../../hypermedia-constants';
+import { performSirenAction } from '../../es6/SirenAction';
+
+export class ContentWorkingCopyEntity extends Entity {
```
brissons, line 5: I saw some discussion by @CodeBaboon, @quinnbudan and @ianlaidlaw (sorry if I missed anyone) about extracting the working copy stuff into its own entity, which can then be extended. Here is a small stab at it. I figure weblinks will eventually extend this class (or something like it) as well.

src/files/FileLocationEntity.js
```diff
@@ -0,0 +1,18 @@
+import { Entity } from '../es6/Entity';
+
+/**
+ * FileLocationEntity class representation of a d2l content file entity.
+ */
+export class FileLocationEntity extends Entity {
```
brissons, line 6: Open to any other name suggestions! The `ContentHtmlFileEntity` has a `getFilesHref()` which, if the link is called, would return this entity. This entity has the link which returns the data of the actual file.
html-sdk
I've been trying to work out how it will look on the activities side, but I believe that once `ContentHtmlFileEntity` is initialized, we would then need to fetch this entity, get the `getFileLocationHref()` then fetch the actual file data.

test/activities/content/ContentHtmlFileEntity.js
quinnbudan: On the one hand it's a little weird that two things can equal while having different descriptions, so maybe it's better to update the equals method, however I agree that as it is now the tests aren't good and if we don't update the equals method I think it makes sense to delete them

src/activities/content/ContentHtmlFileEntity.js
```diff
@@ -0,0 +1,133 @@
...
+		this.dispose();
+	}
+
+	/**
+	 * Checks if content html file properties passed in match what is currently stored
+	 * @param {object} contentHtmlFile Object containing html file specific properties
+	 */
+	equals(contentHtmlFile) {
+		const diffs = [
+			[this.title(), contentHtmlFile.title],
```
quinnbudan, line 113: I don't know how this would be done, but is it at all possible to check links here? I'm just worried that if we have two HTML file entities that have the same title but link to different files we would say they're equal when they aren't

src/activities/content/ContentWorkingCopyEntity.js
quinnbudan: This is great!

test/activities/content/ContentHtmlFileEntity.js
```diff
@@ -0,0 +1,104 @@
...
+		it('Equality should return false when description is different', () => {
+			const htmlFileData = {
+				title: 'Test Html File Title',
+				descriptionRichText: '<p>New description text</p>'
+			};
+			expect(contentHtmlFileEntity.equals(htmlFileData)).to.equal(false);
+		});
+	});
+
+	describe('Actions', () => {
```
quinnbudan, line 67: Nit: I think it might be good to also add a test for the `update-html-content` action

quinnbudan (approved):
LGTM with a few questions/nits!

src/activities/content/ContentHtmlFileEntity.js
brissons: Whoops, definitely missed that. I agree. Thanks!

src/activities/content/ContentWorkingCopyEntity.js
```diff
@@ -0,0 +1,51 @@
...
+			return;
+		}
+
+		return await performSirenAction(this._token, action);
+	}
+
+	/**
+	 * performs a cancel action on the entity
+	 */
+	async cancel() {
```
brissons, line 25: This will be removed.

src/activities/content/ContentHtmlFileEntity.js
```diff
@@ -0,0 +1,133 @@
...
+			return null;
+		}
+		return descriptionSubEntity.properties.text || '';
+	}
+
+	/**
+	 * @returns {string} Title of the content-htmlFile item
+	 */
+	title() {
+		return this._entity && this._entity.properties && this._entity.properties.title;
```
ianlaidlaw, line 37: question about this one here. looking at `descriptionText()` and `descriptionRichText()`, the comment suggests it can return `{string|null}` which makes sense looking at what can happen, but looking at this line here and the fact that it suggests it will always return a string, something doesn't look quite right.

There is checking at each step suggesting that this could short circuit, but there's no fallback or anything. So if `title` isn't there for whatever reason, we're going to return `undefined`. So I would suggest either changing the return type description, or removing the null checking if we're confident it will exist. Or add a fallback empty string like the others.

src/activities/content/ContentHtmlFileEntity.js
```diff
@@ -0,0 +1,133 @@
...
+		}
+		return true;
+	}
+
+	/**
+	* @returns {string} Returns URL of the files home API
+	*/
+	getFilesHref() {
+		if (!this._entity || !this._entity.hasLinkByRel(Rels.Files.file)) {
+			return;
```
ianlaidlaw, line 128: Something similar here. IMO we should be more explicit about what we're returning (maybe return null) instead of an empty return that will get set to `undefined`, but this also doesn't match the expected return type based on the comment above the function

src/activities/content/ContentWorkingCopyEntity.js
ianlaidlaw: love it!

ianlaidlaw (approved):
After my small issues get addressed, LGTM

src/hypermedia-constants.js
```diff
@@ -600,6 +606,7 @@ export const Actions = {
 	},
 	workingCopy: {
 		checkout: 'checkout',
-		checkin: 'checkin'
+		checkin: 'checkin',
```
brissons, line 609: Brain fart this should be `commit`.

test/activities/content/ContentHtmlFileEntity.js
brissons: I agree with you. I will update the equals methods to include both descriptions.

src/activities/content/ContentHtmlFileEntity.js
brissons: I agree with you. I'm going to make a separate PR which will address this and test stuff.

src/activities/content/ContentHtmlFileEntity.js
```diff
@@ -0,0 +1,133 @@
...
+
+		const fields = [{ name: 'title', value: title }];
+		await performSirenAction(this._token, action, fields);
+	}
+
+	/**
+	 * Updates the html file content with the given html
+	 * @param {html} html to set on the html file
+	 */
+	async setHtmlFileHtml(html) {
```
CodeBaboon, line 78: should this be called `setHtmlFileHtmlContent`? That would be more consistent with the action name `htmlContent`.

src/files/FileLocationEntity.js
CodeBaboon: Could it just be a `FileEntity`? It's essentially the siren entity representing the file, which you then use to get the href to the file resource itself, but it could (at least theoretically) have other file metadata on it beyond just the location I think?

src/activities/content/ContentHtmlFileEntity.js
CodeBaboon: Wondering if this should be singular (`getFileHref`) as opposed to plural. The href points to a specific file entry, correct?

CodeBaboon (commented):
Looking pretty good! Just a couple questions/suggestions

test/activities/content/ContentHtmlFileEntity.js
brissons: Looking into this a bit further, this change is a bit more involved since we would also need to change [the corresponding method in activities](https://search.d2l.dev/xref/BrightspaceHypermediaComponents/activities/components/d2l-activity-editor/d2l-activity-content-editor/lti-link/state/content-lti-link.js?r=32a193d5#59) for each activity. Since we are keeping track of what we are using anyways, I'm leaning towards removing the descriptions tests, and we can add them if we decide to include the descriptions in the future.

src/hypermedia-constants.js
brissons: Note: Quiz uses `checkin` so I'll be adding a separate one for `commit`.

CodeBaboon: I wonder if we need to work with them to standardize on a single term (I suspect their `checkin` means the same thing as our `commit`).

src/activities/content/ContentHtmlFileEntity.js
brissons: I was thinking this was static, but after chatting with @ChrisKraljevicD2L, I realized this can change if we change the templates!

PR #307, opened 2021-04-07
US123706 FACE Quizzes: Attempt Condition update values of 0 are handled correctly

mostlyfabulous:
[US123706](https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F478977336380&fdp=true?fdp=true): [ui] Manage Attempts (Attempts allowed, RIO, Attempt conditions)

`0` is also `false` when evaluated in conditionals, this is currently preventing `0` being set as a value for `min` or `max`

### DIFF
--- a/src/activities/quizzes/attempts/QuizAttemptsEntity.js
+++ b/src/activities/quizzes/attempts/QuizAttemptsEntity.js
@@ -166,11 +166,13 @@ export class QuizAttemptsEntity extends Entity {
 	_hasAttemptConditionChanged(attemptCondition) {
 		const original = this.getAttemptConditionSubEntity(attemptCondition.attempt);
 		if (!original || !original.properties) return false;
-		if (!original.properties.min && attemptCondition.min !== undefined) return true;
-		if (!original.properties.max && attemptCondition.max !== undefined) return true;
 		// when `min` or `max` has a value of `undefined` we are deleting the original value for `min` or `max`
-		if ('min' in attemptCondition && original.properties.min && original.properties.min !== attemptCondition.min) return true;
-		if ('max' in attemptCondition && original.properties.max && original.properties.max !== attemptCondition.max) return true;
+		if ('min' in attemptCondition) {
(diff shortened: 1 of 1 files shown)


src/activities/quizzes/attempts/QuizAttemptsEntity.js
```diff
@@ -166,8 +166,8 @@ export class QuizAttemptsEntity extends Entity {
 	_hasAttemptConditionChanged(attemptCondition) {
 		const original = this.getAttemptConditionSubEntity(attemptCondition.attempt);
 		if (!original || !original.properties) return false;
-		if (!original.properties.min && attemptCondition.min) return true;
-		if (!original.properties.max && attemptCondition.max) return true;
+		if (!original.properties.min && attemptCondition.min !== undefined) return true;
```
ronan-f, line 169: What is this checking for? `0`, `NaN` or `null` I suppose? If `min` is `undefined` this condition will evaluate to `false` and if it's `null` or `NaN` it will be `true`? Do we explicitly handle the `undefined` case elsewhere?

mostlyfabulous: If original min is `undefined` and the new min has a numeric value, then it has changed. Perhaps it is safer to be explicit and say: 
`if (original.properties.min === undefined && attemptCondition.min !== undefined) return true;`

ronan-f: Ah ok I was confused about what we were actually checking there. In this case then if `original.properties.min` is `0` and `attemptConditions.min` is also `0` won't it return a false positive?

ronan-f (approved):

src/activities/quizzes/attempts/QuizAttemptsEntity.js
mostlyfabulous: `null` and `NaN` shouldn't be passed into this

ronan-f (approved):
Clean!

PR #308, opened 2021-04-09
US126237: Genericizing html files

brissons:
## Relevant Rally Links 

[US126237](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fuserstory%2F514758865880&fdp=true): BE > HM > Siren SDK updates for content files

## Description

This PR takes the html file entity and genericizes them to generic content files. Everything is _mostly_ the same, except for some minor differences including:
- renaming everything to `ContentFile` from `ContentHtmlFile`
- `getFileType()`: returns what kind of file we have depending on the link class (so far, it only returns `html`; presumably more classes will be added as we add more content files).

## Goal/Solution

Expand the `ContentHmlFileEntity` to support multiple files.

## Related PRs

Quinn has [a great PR here](https://github.com/Brightspace/lms/pull/8481) which genericizes the response, and is what this PR is tailored for.

## Remaining Work
~~- LF feedback on if this approach is 👍~~
~~- Still WIP, so if the BE changes these definitions will need to be updated.~~


src/activities/content/ContentFileEntity.js
```diff
@@ -127,4 +131,17 @@ export class ContentHtmlFileEntity extends ContentWorkingCopyEntity {
 	getFileHref() {
 		return ContentHelperFunctions.getHrefFromRel(Rels.Files.file, this._entity);
 	}
+
+	/**
+	 * @returns {string|null} Returns the File type
+	 */
+	getFileType() {
```
brissons, line 138: The file type is determined by the class in the file link:
```
{
      "class": [
        "file",
        "html"
      ],
      "rel": [
        "https://files.api.brightspace.com/rels/file"
      ],
      "type": "application/vnd.siren+json",
      "href": "https://2864b37c-fb9f-4b5b-ae27-897885ef47ce.files.api.proddev.d2l/6606_645000_6613_File%20(HTML).html/usages/6613"
}
```
...all files will have the `file` class, and the specific file type will be determined by a second class. As of now, we only have the `html` class.

src/activities/content/ContentFileEntity.js
```diff
@@ -1,15 +1,19 @@
...
 import ContentHelperFunctions from './ContentHelperFunctions.js';
 import { ContentWorkingCopyEntity } from './ContentWorkingCopyEntity.js';
 
+export const FILE_TYPES = {
+	html: 'html',
+};
+
 /**
- * ContentHtmlFileEntity class representation of a d2l content-htmlFile entity.
+ * ContentFileEntity class representation of a d2l content-file entity.
```
brissons, line 11: This class still has an html file specific action `setHtmlFileHtmlContent`. If we get any other file entity in the future, they won't have this action, so calling this just won't do anything. As we add more file entities, there may be more methods that do not relate to the specific actions of the file entity.

We could instead have file subclasses for each different file type that would extend this `ContentFileEntity` (ex: remove `setHtmlFileHtmlContent` from `ContentFileEntity`, put it in a `ContentHtmlFileEntity` and have that extend `ContentFileEntity`). So we would need to start with `ContentFileEntity`, find out what type of file it is, then create the specific sub-entity. But I'm not sure if that would be worth it/overengineering (sorry for ramble).

brissons: **Note**: we also don't have access to any other methods before calling the `checkout` action anyways, so having access to callable methods that don't do anything is already kind of present (unless we also added a sub entity with only the `checkout` action, which when called will return an entity which would have to be cast to `ContentFileEntity`, which again might be overkill).

ianlaidlaw: That's a good point. I'm inclined to go down the sub-class path like you mentioned so we can always be confident about what is available to us, which goes along with how hypermedia is supposed to work. It might be over engineering right now, but it's something we keep in mind as we continue to work on the file types.

Looking at things like this:
```
const action = this._entity.getActionByName(Actions.files.deleteFile);
		if (!action) {
			return;
		}
```
seems like something we should avoid doing, possibly. This is supposed to be the Siren representation of this entity and we should know or not if we actually have a certain action, I would think. Maybe this kind of guarding was just for safety though.

src/activities/content/ContentEntity.js
```diff
@@ -3,7 +3,7 @@ import { Rels } from '../../hypermedia-constants';
 import ContentHelperFunctions from './ContentHelperFunctions.js';
 
 export const CONTENT_TYPES = {
-	htmlfile: 'htmlfile',
+	file: 'content-file',
```
ianlaidlaw, line 6: just a general observation: I've noticed sometimes constants are referred to as `file` but sometimes `contentFile`. is there any distinction here, and should we be consistent? for example in this same file:
```
getFileHref() {
		return ContentHelperFunctions.getHrefFromRel(Rels.Content.fileEntity, this._entity);
	}
```
should this be `getContentFileHref()`?

brissons: I see what you mean. I'll do a quick scan and see if there are any places where it should be `contentFile`.

src/activities/content/ContentFileEntity.js
brissons: I think we'll still need this guarding down the line, since a lot of actions won't be available if working copy is enabled and it hasn't been checked in yet. However, yes it would be weird if we have a pdf file and there is a `changeHtmlContent` method (or something) available. I'll pursue the sub-class path.

ianlaidlaw: makes sense!

src/activities/content/ContentFileEntity.js
```diff
@@ -1,15 +1,19 @@
-import { Actions, Rels } from '../../hypermedia-constants.js';
+import { Actions, Rels, Classes } from '../../hypermedia-constants.js';
 import { performSirenAction } from '../../es6/SirenAction.js';
 import ContentHelperFunctions from './ContentHelperFunctions.js';
 import { ContentWorkingCopyEntity } from './ContentWorkingCopyEntity.js';
 
+export const FILE_TYPES = {
```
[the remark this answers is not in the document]
brissons: On the surface I like the idea of keeping the constants with the file, similar to how we have `CONTENT_TYPES` in `ContentEntity.js`, and both are imported in the same line in `activities`. Is there a danger if we export constants? If this is a bad pattern though, I can make a new file and move this there.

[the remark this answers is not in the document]
ianlaidlaw: that does make sense, I don't want to mess with the existing pattern. It just depends where these constants will be used, but if they're always used with the component then it's probably fine.

quinnbudan (approved):
Looks great!

CodeBaboon (approved):
LGTM

ianlaidlaw (approved):
LGTM!!

PR #309, opened 2021-04-09
US125502 some initial boilerplate before changing existing api calls …

mdulat:
…to associate grades.


wongvincent (approved):
One thing I worry about is if they switch between the two radio buttons of "create new" and "Link to existing" quickly and back and forth, the UI can become out of sync.

And we haven't thought about doing optimistic UI rendering which means if we click "Create new", the "add scheme" button isn't going to work properly until it comes back with the `get-schemes` endpoint. The FACE quiz timing/attempts does optimistic UI rendering.

This probably belongs in a slack convo. @mdgbayly

wongvincent:
LGTM!

PR #310, opened 2021-04-09
Us126777/add existing categories to UI

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F519563556284

Added methods to siren-sdk for interacting with categories.

### DIFF
--- a/src/activities/CategoriesEntity.js
+++ b/src/activities/CategoriesEntity.js
@@ -3,7 +3,7 @@ import { Classes, Actions } from '../hypermedia-constants.js';
 // import { performSirenAction } from '../es6/SirenAction';
 
 /**
- * CategoriesEntity class representation of an attachment (link or file)
+ * CategoriesEntity class representation
  */
 export class CategoriesEntity extends Entity {
 
@@ -30,7 +30,7 @@ export class CategoriesEntity extends Entity {
 			return;
 		}
 
-		return !!this._entity.getActionByName(Actions.assignments.add);
+		return this._entity.hasActionByName(Actions.assignments.add);
 	}
 
 	/**
--- a/src/activities/CategoriesEntity.js
+++ b/src/activities/CategoriesEntity.js
@@ -1,5 +1,5 @@
-import { Entity } from '../es6/Entity';
-import { Classes } from '../hypermedia-constants.js';
+import { Entity } from '../es6/Entity.js';
+import { Classes, Actions } from '../hypermedia-constants.js';
 // import { performSirenAction } from '../es6/SirenAction';
 
 /**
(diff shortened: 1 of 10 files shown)


src/activities/CategoriesEntity.js
```diff
@@ -0,0 +1,53 @@
+import { Entity } from '../es6/Entity.js';
+import { Classes, Actions } from '../hypermedia-constants.js';
+// import { performSirenAction } from '../es6/SirenAction';
+
+/**
+ * CategoriesEntity class representation of an attachment (link or file)
```
mdulat, line 6: Probably not of an attachment :)

src/activities/CategoriesEntity.js
```diff
@@ -0,0 +1,53 @@
...
+
+	/**
+	 * @returns {Boolean} Whether or not a user can edit categories
+	*/
+	canEditCategories() {
+		if (!this._entity) {
+			return;
+		}
+
+		return !!this._entity.getActionByName(Actions.assignments.add);
```
mdulat, line 33: can you use `hasActionByName` instead?

test/activities/CategoriesEntity.js
```diff
@@ -0,0 +1,88 @@
...
+			const selected = categoriesEntity.getSelectedCategory();
+
+			expect(selected).to.be.undefined;
+		});
+
+		it('GetsSelectedCategoriesWithNoActions', () => {
+			var categoriesEntity = new CategoriesEntity(nonEditableEntity);
+			const selected = categoriesEntity.getSelectedCategory();
+
+			expect(selected).to.be.undefined;
```
mdulat, line 67: Can't a non-editable entity still have a selected class? I.e. should the UI be able to see that an assignment has a category (if it does) even if we can't modify it?

test/activities/CategoriesEntity.js
```diff
@@ -0,0 +1,88 @@
...
+		});
+	});
+
+	describe('Editable', () => {
+		it('sets canEditName to true', () => {
```
mdulat, line 72: `canEditCategories`? (not Name)

test/activities/CategoriesEntity.js
```diff
@@ -0,0 +1,88 @@
...
+			var categoriesEntity = new CategoriesEntity(editableEntity);
+
+			const res = categoriesEntity.canEditCategories();
+			expect(res).to.be.true;
+		});
+	});
+
+	describe('non editable', () => {
+		it('sets canEditName to true', () => {
```
mdulat, line 81: s.b. "sets canEditCategories to false"

test/activities/CategoriesEntity.js
```diff
@@ -57,19 +57,21 @@ describe('CategoriesEntity', () => {
 			var categoriesEntity = new CategoriesEntity(nonEditableEntity);
```
mdulat, line 57: Hmm, I think for this test "when not present" we should have an entity that actually has no category selected, and expect `selected` to be undefined then.

test/activities/CategoriesEntity.js
```diff
@@ -78,7 +80,7 @@ describe('CategoriesEntity', () => {
 	});
 
 	describe('non editable', () => {
-		it('sets canEditName to true', () => {
+		it('sets canEditCategories to true', () => {
```
mdulat, line 83: Still need to update `true` to `false` in this test name, based on what it's doing.

mdulat (approved):

PR #313, opened 2021-04-14
Add a lesson-view-page alt html link to activity-usage entities for content topics

CodeBaboon:
## Relevant Rally Links 

[US126672](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F519007142480&fdp=true?fdp=true): Modify activities saveHref to handle new working copy items

## Description

Adds a method to retrieve the `lesson-view-page` link from the `ContentEntity`, as well as the `activitiy-usage` link from the `ContentWebLink` entity.

## Goal/Solution

The `lesson-view-page` link will allow the FACE UI to navigate back to the correct topic after saving (especially useful for creating a new topic).

The `activity-usage` link allows the web link to raise an event with the updated activity-usage href value when saving. This is important when committing a new unsaved weblink for the first time as the activity usage href will change and the parent control needs to be aware so it can fetch this updated entity.

## Video/Screenshots

n/a

### DIFF
--- a/src/activities/content/ContentEntity.js
+++ b/src/activities/content/ContentEntity.js
@@ -39,35 +39,35 @@ export class ContentEntity extends Entity {
 	}
 
 	/**
-	 * @returns {string} content-module link
+	 * @returns {string|null} content-module link
 	 */
 	getModuleHref() {
 		return ContentHelperFunctions.getHrefFromRel(Rels.Content.moduleEntity, this._entity);
 	}
 
 	/**
-	 * @returns {string} content-weblink link
+	 * @returns {string|null} content-weblink link
 	 */
 	getWebLinkHref() {
 		return ContentHelperFunctions.getHrefFromRel(Rels.Content.weblinkEntity, this._entity);
 	}
 
 	/**
-	 * @returns {string} content-ltilink link
+	 * @returns {string|null} content-ltilink link
 	 */
 	getLTILinkHref() {
 		return ContentHelperFunctions.getHrefFromRel(Rels.Content.ltilinkEntity, this._entity);
 	}
 
 	/**
-	 * @returns {string} content-htmlfile link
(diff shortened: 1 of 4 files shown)


quinnbudan (approved):

brissons (approved):
lgtm!

src/activities/content/ContentEntity.js
```diff
@@ -65,4 +65,12 @@ export class ContentEntity extends Entity {
 	getHtmlFileHref() {
 		return ContentHelperFunctions.getHrefFromRel(Rels.Content.htmlFileEntity, this._entity);
 	}
+
+	/**
+	 * @returns {string} lesson-view-page link
```
ianlaidlaw, line 70: Am i right to assume that if this link won't always be added then the return type would be `{string|undefined}` (or null?) ? I know we've updated that kind of stuff in other spots but not here.

ianlaidlaw: It's not a big deal since we don't really _need_ this "typing". but to me I'd rather be accurate or just not even include it at all.

ianlaidlaw (approved):

src/activities/content/ContentEntity.js
CodeBaboon: You are correct, I just went with the typing of the other functions using the same helper. I think it can also return `null` (rather than `undefined`). Here is the implementation of the helper function:

```
getHrefFromRel: (rel, entity) => {
		return entity
			&& entity.hasLinkByRel(rel)
			&& entity.getLinkByRel(rel).href;
	}
```

CodeBaboon: Resolved in https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/313/commits/56f165dfdefe5ef26b2d4ba4d399f39869804d18

PR #315, opened 2021-04-14
feat: US125502 add set grade functions and unit tests

mdulat:
** DO NOT MERGE YET **

Just adding some functions and unit tests.


wongvincent (commented):
looking good!

src/activities/associateGrade/AssociateGradeEntity.js
```diff
@@ -1,22 +1,66 @@
...
 export class AssociateGradeEntity extends Entity {
+
 	canEditGradebookStatus() {
 		return this._entity && this._entity.hasActionByName(Actions.activities.associateGrade.gradebookStatus);
 	}
 
-	async setGradebookStatus(newStatus) {
+	canEditNewGrade() {
+		const newGradeEntity = this._getNewGradeEntity();
+		return typeof newGradeEntity !== 'undefined' && newGradeEntity.hasActionByName(Actions.activities.associateGrade.chooseType);
```
wongvincent, line 20: can just be `return newGradeEntity&& newGradeEntity.hasActionByName(Actions.activities.associateGrade.chooseType);`?

wongvincent (approved):

src/activities/associateGrade/AssociateGradeEntity.js
mdulat: Nope because it has to return `true` or `false`, not `undefined`. I had what you suggest initially, which failed tests that expect `false`.

PR #317, opened 2021-04-15
Us126389/add categories

ronan-f:
https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F516602039000


src/activities/CategoriesEntity.js
```diff
@@ -70,31 +70,49 @@ export class CategoriesEntity extends Entity {
...
 
-		const action = categoryEntity.getActionByName(Actions.assignments.categories.select);
-		if (!action) {
-			return;
+		if (category.categoryName) {
+			actions.push(this._generateNewCategoryAction(category.categoryName));
 		}
 
-		await performSirenAction(this._token, action);
+		await Promise.all(actions.map(({ action, fields }) => performSirenAction(this._token, action, fields)));
```
mdulat, line 116: Are we ever going to be executing more than one of these actions per save? Aren't we either setting it to no category, setting it to a given category, or creating a new category? 

Also, if this is just a PATCH on the assignments endpoint, can we combine this update with other updates on the same endpoint using `performSirenActions`, similarly to [how it's done for ActivityUsage](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/ActivityUsageEntity.js#L674)?

ronan-f: > Are we ever going to be executing more than one of these actions per save? Aren't we either setting it to no category, setting it to a given category, or creating a new category?

True it should only be one action at a time. 

> Also, if this is just a PATCH on the assignments endpoint

Hmm so the `select`/`deselect` actions are on the PATCH assignments endpoint whereas adding a new category is a POST to /categories. I was trying to make it work with `performSirenActions` but it don't know that it supports multiple endpoints?

mdulat: It does support multiple endpoints, it will [combine actions that share the same href](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/es6/SirenAction.js#L201). Though I'm noticing now that `CategoriesEntity` isn't within the `assignments` subfolder, so assuming this may be relevant to activities other than assignments? If so then other activities would use different endpoints. We'd need to combine the actions in the CategoriesEntity with those in AssignmentEntity for assignment saves, and with other entities for other activity type saves.

mdulat (approved):

PR #318, opened 2021-04-28
Added new method to check if a user can add categories

ronan-f:
Part of the fix for [this trello card](https://trello.com/c/PPE1Nhce/364-categories-if-user-has-add-edit-assignment-submission-folders-permission-they-see-read-only-view-they-should-have-the-ability-to).

In the initial implementation I was only using `canEditCategories` to determine a user's interaction with categories but in fact there is a need to check if they can edit categories AND if they can add new categories.

### DIFF
--- a/src/activities/assignments/CategoriesEntity.js
+++ b/src/activities/assignments/CategoriesEntity.js
@@ -33,14 +33,26 @@ export class CategoriesEntity extends Entity {
 	}
 
 	/**
-	 * @returns {Boolean} Whether or not a user can edit categories
+	 * @returns {Boolean} Whether or not a user can add categories
 	*/
-	canEditCategories() {
+	canAddCategories() {
 		if (!this._entity) {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


src/activities/assignments/CategoriesEntity.js
```diff
@@ -33,16 +33,28 @@ export class CategoriesEntity extends Entity {
...
+	/**
+	 * @returns {Boolean} Whether or not a user can edit categories
+	*/
+	canEditCategories() {
+		const categories = this.getCategories();
+		if (!categories || !categories.length) {
+			return;
+		}
+
+		return categories[0].hasActionByName(Actions.assignments.categories.select);
```
mdulat, line 55: Is it possible that the first category will be selected, in which case we would be looking for the `deselect` action?

ronan-f: That's a very good point. Changing it to check for both.

mdulat (approved):

PR #319, opened 2021-05-04
Update categories to handle new permissions checks

ronan-f:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F516602039000
https://trello.com/c/3nsAWve6/372-when-a-course-has-no-categories-categories-cannot-be-added-in-face

Handling new categories permissions by checking for the "collection" class. Also cleaned up how dirty check is handled here.

### DIFF
--- a/src/activities/assignments/CategoriesEntity.js
+++ b/src/activities/assignments/CategoriesEntity.js
@@ -72,19 +72,25 @@ export class CategoriesEntity extends Entity {
 		return selectedCategory;
 	}
 
+	getSelectedCategoryName() {
+		if (!this._entity) {
+			return;
+		}
+
+		return this._entity.properties.selectedCategory;
+	}
+
 	equals(category) {
 		const selectedCategory = this.getSelectedCategory();
 		if (!selectedCategory) {
 			return category.categoryId === selectedCategory;
 		}
 
-		return category.categoryId === selectedCategory.properties.categoryId;
+		return Number(category.categoryId) === Number(selectedCategory.properties.categoryId);
 	}
 
-	_generateDeselectCategoryAction(category) {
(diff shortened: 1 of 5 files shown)


emil-furniss:
This seems like an odd approach to handle a permissions. Normally we do that by in/excluding actions, no? Looking at the current state it looks to me like a UI bug. The API contains the correct `AddCategory` action but it's just not checking it. I might be missing something, but looks to me like [this activities code](https://github.com/BrightspaceHypermediaComponents/activities/blob/aa4f875037b161d16721206e6f26c386693e31ea/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-categories-editor.js#L69) should also be checking [this permission](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/b71323f397a256423e4e1f28e59d94d1bcc73479/src/activities/assignments/CategoriesEntity.js#L38) from the siren-sdk.
The render method might need adjusting as well, but that seems like the jist of things?

src/activities/assignments/CategoriesEntity.js
```diff
@@ -110,16 +116,23 @@ export class CategoriesEntity extends Entity {
...
+	_hasCategoryIdChanged(categoryId) {
+		const selectedCategory = this.getSelectedCategory();
+		const initialId = selectedCategory && selectedCategory.categoryId;
+
+		return categoryId !== initialId;
+	}
+
 	async save(category) {
-		if (!this.canEditCategories()) return;
+		if (!this.canAddCategories()) return;
```
mdulat, line 127: Is this right? How do they select a new category when they don't have addCategory permissions then?

mdulat (approved):

dustxd (approved):

PR #320, opened 2021-05-07
US116840: Add method to retrieve activity-usage link from ContentFileEntity

brissons:
## Relevant Rally Links 
[US116840](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fuserstory%2F393033401500&fdp=true): FACE HTML file - Add HTML field via new HTML editor
[US126672](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F519007142480&fdp=true?fdp=true): Modify activities saveHref to handle new working copy items

## Description
Adding method to retrieve the `activity-usage` link from the ContentFileEntity.

## Goal/Solution
Similar to [this PR](https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/313), we will need the `activity-usage` link in the content file entity in order to get the updated `lesson-view-page` link, which allows the FACE UI to navigate back to the correct topic in lessons after saving.

### DIFF
--- a/src/activities/content/ContentFileEntity.js
+++ b/src/activities/content/ContentFileEntity.js
@@ -35,7 +35,7 @@ export class ContentFileEntity extends ContentWorkingCopyEntity {
 	}
 
 	/**
-	 * @returns {string|null} activity usage link
+	 * @returns {string|null} Activity usage link
 	 */
 	getActivityUsageHref() {
 		return ContentHelperFunctions.getHrefFromRel(Rels.Activities.activityUsage, this._entity);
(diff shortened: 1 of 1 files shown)


quinnbudan (approved):
Looks good! But I think I'm confused: why do we need the activity usage link to get the lessons view link? Is the lessons view link only on the usage and not the content entity itself?

brissons:
> Looks good! But I think I'm confused: why do we need the activity usage link to get the lessons view link? Is the lessons view link only on the usage and not the content entity itself?

Yup that is correct! [This link right here](https://user-images.githubusercontent.com/52549862/117481397-8f243700-af28-11eb-9278-a0da9a46dfbe.png).


We need the `activity-usage`  to get the updated `lessons-view-page` from the activity once its been saved in order to redirect to the correct topic in lessons.

PR #323, opened 2021-05-13
US128071 - get file type class from entity

quinnbudan:
## Relevant Rally Links 

- [US128071](https://rally1.rallydev.com/#/289692574792d/dashboard?detail=%2Fuserstory%2F604152072952&fdp=true?fdp=true): (FE) FACE HTML file - Don't load file usage link if not present, change where to check for type

## Description

To handle the fact that we don't want to have a file entity usage link when the file activity is new, we have removed the link from the entity response when it's a new entity. Because of this, we can't rely on the file-type class being on this link, so we've moved it to the entity itself. The SDK needs to be updated to look in the right spot for the file-type class.

## Goal/Solution

The update here is to check the entity's classes rather than the link's classes for a matching type. I also updated this to use a map rather than a series of if-statements. This might sacrifice some readability but will stop the check from becoming huge and allow us to add other class-type mappings in the future as one-liners. I'm not sure that this tradeoff is worth it, so I'm happy to go back to an if-statement or a switch-statement!

I don't know what's going on with the whitespace changes, the spacing is still tabs and line endings didn't change.

## Video/Screenshots

N/A

## Related PRs

- https://github.com/BrightspaceHypermediaComponents/activities/pull/1727 the activities PR can be merged after this PR goes in

## Remaining Work

N/A

### DIFF
--- a/src/activities/content/ContentFileEntity.js
+++ b/src/activities/content/ContentFileEntity.js
@@ -1,10 +1,14 @@
-import { Actions, Rels, Classes } from '../../hypermedia-constants.js';
+import { Actions, Rels } from '../../hypermedia-constants.js';
 import { performSirenAction } from '../../es6/SirenAction.js';
 import ContentHelperFunctions from './ContentHelperFunctions.js';
 import { ContentWorkingCopyEntity } from './ContentWorkingCopyEntity.js';
 
 export const FILE_TYPES = {
 	html: 'html',
+};
+
+const CLASS_FILE_TYPE_MAP = {
+	html: FILE_TYPES.html,
 };
 
 /**
@@ -126,12 +130,15 @@ export class ContentFileEntity extends ContentWorkingCopyEntity {
 	 * @returns {string|null} Returns the File type
 	 */
 	getFileType() {
-		if (!this._entity && !this._entity.hasLinkByClass(Classes.files.file)) {
+		if (!this._entity) {
 			return null;
-		} else if (this._entity.hasLinkByClass(Classes.files.html)) {
-			return FILE_TYPES.html;
 		}
 
+		for (const [key, value] of Object.entries(CLASS_FILE_TYPE_MAP)) {
+			if (this._entity.hasClass(key)) {
+				return value;
+			}
+		}
 		return null;
 	}
 }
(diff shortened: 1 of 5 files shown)


src/activities/content/ContentFileEntity.js
```diff
@@ -1,137 +1,144 @@
...
+import { Actions, Rels } from '../../hypermedia-constants.js';
+import { performSirenAction } from '../../es6/SirenAction.js';
+import ContentHelperFunctions from './ContentHelperFunctions.js';
+import { ContentWorkingCopyEntity } from './ContentWorkingCopyEntity.js';
+
+export const FILE_TYPES = {
+	html: 'html',
+};
+
+const CLASS_FILE_TYPE_MAP = {
```
quinnbudan, line 10: This is the map that I added, in the future we can just add new types here as one line of code and everything will work. Again there's some tradeoffs to this solution that I mentioned in my PR description so I'm happy to abandon this and go back to an if-statement / switch-statement!

src/activities/content/ContentFileEntity.js
```diff
@@ -1,137 +1,144 @@
...
+
+	/**
+	 * @returns {string|null} Returns the File type
+	 */
+	getFileType() {
+		if (!this._entity) {
+			return null;
+		}
+
+		for (const [key, value] of Object.entries(CLASS_FILE_TYPE_MAP)) {
```
quinnbudan, line 137: The biggest change here is that we now use the map and do the `.hasClass` right on the entity rather than checking the link.

quinnbudan (commented):
I'm doing a bit of the tour because of the whitespace issues

test/activities/content/ContentFileEntity.js
```diff
@@ -1,100 +1,102 @@
...
+		});
+	});
+
+	describe('Links', () => {
+		it('can get file href', () => {
+			expect(contentFileEntity.getFileHref()).to.equal('https://fake-tenant-id.files.api.proddev.d2l/my-file.file/usages/6614');
+		});
+	});
+
+	describe('Classes', () => {
```
quinnbudan, line 97: I also moved the file type test to here

src/activities/content/ContentFileEntity.js
CodeBaboon: Love this map! I think this is definitely the way to go.

CodeBaboon (approved):
Looks good!

src/activities/content/ContentFileEntity.js
ianlaidlaw: yep, much better

brissons (approved):
lgtm!

PR #326, opened 2021-05-26
Fix dirty check

wongvincent:
https://trello.com/c/QM0Z876K/335-face-quiz-dirty-check-dialog-is-triggered-for-new-quiz

When scoreOutOf is 0, we convert it to `''` an empty string. So this comparison was failing because it would compare "0" with ''.

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -688,9 +688,9 @@ export class ActivityUsageEntity extends Entity {
 		];
 
 		// For activities that don't use scoreOutOf, this.scoreOutOf() is initialized to undefined
-		const scoreOutOf = `${this.scoreOutOf()}`;
+		const scoreOutOf = this.scoreOutOf();
 		if (scoreOutOf !== undefined) {
-			diffs.push([scoreOutOf, activity.scoreAndGrade.scoreOutOf]);
+			diffs.push([String(scoreOutOf), activity.scoreAndGrade.scoreOutOf]);
 		}
 
 		for (const [left, right] of diffs) {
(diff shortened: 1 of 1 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -687,10 +687,9 @@ export class ActivityUsageEntity extends Entity {
 			[this.inGrades(), activity.scoreAndGrade.inGrades]
 		];
 
+		// For activities that don't use scoreOutOf, this.scoreOutOf() is initialized to undefined
 		const scoreOutOf = `${this.scoreOutOf()}`;
-
-		// For activities that don't use scoreOutOf, this.scoreOutOf() is initialized to undefined whereas for those that do it initializes to ''
-		if (scoreOutOf !== 'undefined') {
+		if (scoreOutOf !== undefined) {
```
ronan-f, line 692: Is this check correct? Will `scoreOutOf` ever be `undefined` here? Regardless of how it's initialized it will be parsed as a string on 691 right? 

I'm thinking of the case when `this.scoreOutOf()` returns `undefined`. We'll be comparing the string `'undefined'` with the value `undefined` and end up with the string added to the diffs. Is that intended?

wongvincent: When `this.scoreOutOf()` is called, doesn't it possibly return `undefined` and not the string `"undefined"`?
That's why I changed this to check for the non-string version.

ronan-f: Yeah but even if the function call returns the value `undefined` we are then parsing it as a string by using the template literal syntax on line 691

image

ronan-f: imo intent would be clearer doing something like this:

```js
const scoreOutOf = this.scoreOutOf();
if (scoreOutOf !== undefined) {
  diffs.push([String(scoreOutOf), activity.scoreAndGrade.scoreOutOf]);
}
```
so we actually check the value returned from the function and not the stringified version and then explicitly cast it to a string when adding to diffs 🤷

wongvincent: yes

ronan-f (approved):
gg

PR #328, opened 2021-06-01
Mdulat/us125502 get categories

mdulat:
Adds functionality for getting grade categories

### DIFF
--- a/src/activities/associateGrade/AssociateGradeEntity.js
+++ b/src/activities/associateGrade/AssociateGradeEntity.js
@@ -89,7 +89,7 @@ export class AssociateGradeEntity extends Entity {
 		// HACK adding query param as field due to bug in performSirenAction (_getSirenFields function)
 		const url = new URL(action.href, window.location.origin);
 		const wcId = url.searchParams.get('workingCopyId');
-		fields.push({ name: 'workingCopyId', value: wcId } );
+		fields.push({name: 'workingCopyId', value: wcId});
 
 		const returnedEntity = await performSirenAction(this._token, action, fields);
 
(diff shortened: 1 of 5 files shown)


src/activities/associateGrade/AssociateGradeEntity.js
```diff
@@ -68,11 +69,34 @@ export class AssociateGradeEntity extends Entity {
 		return typeof newGradeEntity !== 'undefined' && newGradeEntity.hasActionByName(Actions.activities.associateGrade.chooseType);
 	}
 
+	canGetCategories() {
+		const newGradeEntity = this._getNewGradeEntity();
+		return typeof newGradeEntity !== 'undefined' && newGradeEntity.hasActionByName(Actions.activities.associateGrade.getCategories);
```
wongvincent, line 74: any reason we don't just do `return newGradeEntity && newGradeEntity.hasActionByName(Actions.activities.associateGrade.getCategories);`

src/activities/associateGrade/GradeCategoryEntity.js
```diff
@@ -0,0 +1,13 @@
+import { Entity } from '../../es6/Entity';
+
+/**
+ * GradeCategoryEntity class representation of a grade category
+ */
+export class GradeCategoryEntity extends Entity {
```
wongvincent, line 6: Could we use the `GradeCategoryEntity` in `src/activities` instead of a new one that does the same thing. I think they'll always be the same.

wongvincent (approved):

src/activities/associateGrade/AssociateGradeEntity.js
mdulat: yeah, it's so that `canGetCategories` always returns true or false, not undefined. This is necessary for tests, which reminds me I need to add tests for these new functions.

src/activities/associateGrade/GradeCategoryEntity.js
mdulat: They are the same, my thinking is instead of having two locations for these related entities we would put this into the `associateGrade` folder, then when we are no longer using the original one we can remove it from `src/activities` (I guess we don't have to wait, we can just have activities use the new location right away).

PR #340, opened 2021-06-16
Mdulat/us125503 add schemes dropdown

mdulat:
Adds `isDefault` to scheme entity and returns `AssociateGradeEntity` after selecting scheme.

https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fuserstory%2F505558536824&fdp=true&itemRef=%2Fhierarchicalrequirement%2F505558536824&previousTokens=%2F29180338367ud%2Fcustom%2F486232622040%3Fdetail%3D%252Fuserstory%252F505558536824&project=%2Fproject%2F29180338367&ref=%2Fhierarchicalrequirement%2F505558536824&relationship=tasks

### DIFF
--- a/src/activities/associateGrade/GradeSchemeLinkedEntity.js
+++ b/src/activities/associateGrade/GradeSchemeLinkedEntity.js
@@ -1,3 +1,4 @@
+import { AssociateGradeEntity } from './AssociateGradeEntity';
 import { Entity } from '../../es6/Entity.js';
 import { Classes, Actions, Rels } from '../../hypermedia-constants.js';
 import { performSirenAction } from '../../es6/SirenAction.js';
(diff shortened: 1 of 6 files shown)


src/activities/associateGrade/GradeSchemeLinkedEntity.js
```diff
@@ -28,12 +29,27 @@ export class GradeSchemeLinkedEntity extends Entity {
...
+	isDefault() {
+		if (!this._entity) {
+			return false;
+		}
+		return this._entity.hasClass(Classes.activities.associateGrade.default);
+	}
+
 	async selectScheme() {
-		if (!this.canChooseScheme()) {
+		if (!this.canChooseScheme() || this.isSelected()) {
```
wongvincent, line 40: any chance each `GradeSchemeLinkedEntity` isn't updated so the entity may have the `selected` class, but not actually be the selected item in the UI?
Do we update the collection of grade schemes?
https://github.com/BrightspaceHypermediaComponents/activities/blob/v3.179.0/components/d2l-activity-editor/state/associate-grade.js#L45
Should that be called in `load`?
I also tried to optimize it so the categories and candidates aren't constantly re-fetched when the `associate-grade` entity is updated by doing the `this.gradeCategoryCollection || this.getGradeCategories();` thing. If schemes also did that, I think the collection might not update?

wongvincent: I see your activities repo changes. They look good!

wongvincent (approved):

PR #341, opened 2021-06-16
US129035 - Prevent error thrown when category has no href

wongvincent:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F601489857733&fdp=true?fdp=true


mdulat (approved):
Just wondering when would this be the case? Doesn't a category linked entity always have a category link?

mdulat:
> Just wondering when would this be the case? Doesn't a category linked entity always have a category link?

Never mind, I see why in your lms PR.

PR #337, opened 2021-06-11
US128812: LTI - Add LinkFrameOptionsEntity

brissons:
## Relevant Rally Links

[US128812](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fuserstory%2F600911792617&fdp=true): LTI page - Add link to LTI link entity that will check X-Frame-Options header

## Description

Adding a `siren-sdk` definition for the `frame-options` entity for LTI links.

## Goal/Solution

We will use this definition in activities to check if we should allow for embedding, and if we should try to render an embedded preview. A more detailed explanation can be found in the related lms PR.

## Related PRs

https://github.com/Brightspace/lms/pull/11335

PR #343, opened 2021-06-16
DE43998: [cert] LTI - Add LinkFrameOptionsEntity

brissons:
## Relevant Rally Links

[DE43998](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fdefect%2F601841380817&fdp=true): [cert] LTI page - Add link to LTI link entity that will check X-Frame-Options header

## Description

Backport of https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/337 for 20.21.7.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - master
+      - '[0-9]+.x'
+      - '[0-9]+.[0-9]+.x'
 jobs:
   release:
     if: "!contains(github.event.head_commit.message, 'skip ci')"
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - master
+      - '[0-9]+.x'
```
ianlaidlaw, line 6: should these be added to master as well?

ianlaidlaw (approved):

.github/workflows/release.yml
brissons: I think it probably should. I can make a different PR to add it.

PR #358, opened 2021-06-30
US127570 - update template entity definition to provide a title function

quinnbudan:
## Relevant Rally Links

- [US127570](https://rally1.rallydev.com/#/289692574792d/dashboard?detail=%2Fuserstory%2F602050357608&fdp=true?fdp=true): FACE HTML file - Support Sorting HTML Templates By Name

## Description

We need access to the HTML template entity's title.

## Goal/Solution

Provide a `title()` function on the entity that returns the title and does appropriate null checking.

## Video/Screenshots

N/A

## Related PRs

- https://github.com/BrightspaceHypermediaComponents/activities/pull/1860

## Remaining Work

If applicable...

- [ ] Dev Testing: another developer will test this code on their machine

### DIFF
--- a/src/activities/content/ContentHtmlFileTemplatesEntity.js
+++ b/src/activities/content/ContentHtmlFileTemplatesEntity.js
@@ -4,14 +4,6 @@ import { Entity } from '../../es6/Entity';
  * ContentHtmlFileTemplatesEntity class representation of an html file templates entity.
  */
 export class ContentHtmlFileTemplatesEntity extends Entity {
-
-	/**
-	 * @returns {string|undefined} Title of the content HTML file template entity
-	 */
-	title() {
-		return this._entity && this._entity.properties && this._entity.properties.title;
-	}
-
 	/**
 	 * @returns {array|undefined} Array containing the subentites of the file template entity, if they exist
 	 */
--- a/src/activities/content/ContentHtmlFileTemplatesEntity.js
+++ b/src/activities/content/ContentHtmlFileTemplatesEntity.js
@@ -6,6 +6,13 @@ import { Entity } from '../../es6/Entity';
 export class ContentHtmlFileTemplatesEntity extends Entity {
 
 	/**
+	 * @returns {string|undefined} Title of the content HTML file template entity
+	 */
+	title() {
+		return this._entity && this._entity.properties && this._entity.properties.title;
+	}
+
+	/**
 	 * @returns {array|undefined} Array containing the subentites of the file template entity, if they exist
 	 */
 	getHtmlFileTemplates() {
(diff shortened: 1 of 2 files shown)


src/activities/content/ContentHtmlFileTemplatesEntity.js
```diff
@@ -1,14 +1,21 @@
...
+
+/**
+ * ContentHtmlFileTemplatesEntity class representation of an html file templates entity.
+ */
+export class ContentHtmlFileTemplatesEntity extends Entity {
+
+	/**
+	 * @returns {string|undefined} Title of the content HTML file template entity
+	 */
+	title() {
```
quinnbudan, line 11: I'm not sure why there are all these whitespace changes, this should be the only changes (I have LF line endings)

CodeBaboon (approved):

coffeexcode (approved):
LGTM

brissons (approved):

ianlaidlaw (approved):

prajapac (approved):

src/activities/content/ContentHtmlFileTemplatesEntity.js
ianlaidlaw: i did a look at some of our files in this repo and it's a mix of CRLF and LF, I'm guessing this repo doesn't have a linter rule for it, so it's probably just using who ever made the file and if someone with a different setting edits it, if their editor has it set to, it will overwrite that file. but yeah, seems like there isn't a standard set, even looking at some files that don't belong to us

src/activities/content/ContentHtmlFileTemplatesEntity.js
```diff
@@ -1,14 +1,13 @@
...
-export class ContentHtmlFileTemplatesEntity extends Entity {
-
-	/**
-	 * @returns {array|undefined} Array containing the subentites of the file template entity, if they exist
-	 */
-	getHtmlFileTemplates() {
-		return this._entity && this._entity.getSubEntitiesByRel('about');
-	}
-}
+import { Entity } from '../../es6/Entity';
```
quinnbudan, line 1: Just whitespace changes here

src/files/FileEntity.js
```diff
@@ -1,29 +1,35 @@
...
+import { Entity } from '../es6/Entity';
+
+/**
+ * FileEntity class representation of a d2l content file entity.
+ */
+export class FileEntity extends Entity {
+	/**
+	 * @returns {string|undefined} Title of the content HTML file template entity
+	 */
+	title() {
```
quinnbudan, line 10: This is mostly whitespace, but this is the real change. I had to move the `title` to here because the sub-entities on the template entity are `FileEntity` objects, not `TemplateEntity` objects

CodeBaboon (approved):

prajapac (approved):

PR #359, opened 2021-07-07
Stop sending Assignment Patch when nothing has changed

wongvincent:
https://trello.com/c/A61OrkUc/296-on-assignment-save-patch-api-on-folder-is-always-executed-even-when-there-are-no-changes-request-body-includes-notificationemail

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -803,16 +803,16 @@ export class AssignmentEntity extends Entity {
 			this.canEditAnnotations();
 
 		if (shouldSaveAnnotations) {
-
 			fields.push({
 				name: 'annotationToolsAvailability',
 				value: assignment.annotationToolsAvailable
 			});
 		}
 
-		if (typeof assignment.submissionType !== 'undefined' &&
-				assignment.submissionType !== this.submissionType() &&
-				this.canEditSubmissionType()) {
+		const shouldSaveSubmissionType = typeof assignment.submissionType !== 'undefined' &&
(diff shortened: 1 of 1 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -829,8 +829,8 @@ export class AssignmentEntity extends Entity {
 		}
 
 		if (typeof assignment.completionType !== 'undefined' &&
-				assignment.completionType !== this.completionType() &&
-				this.canEditCompletionType()) {
+			(shouldSaveSubmissionType || assignment.completionType !== this.completionTypeValue()) &&
```
wongvincent, line 832: If `submissionType` is included in the PATCH, we must include the `completionType` (even if it hasn't changed), otherwise it will be reset to `onSubmission`.
https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/Service/Default/DropboxService.cs?r=a46e549b#1149

pureooze: Is it possible for the user to have `canEditSubmissionType` but not have `canEditCompletionType`?

wongvincent: Not possible with the way we are returning the siren entity. Either both actions are included or they're both not included.

https://search.d2l.dev/xref/lms/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs?r=f4e056be#1261-1277

pureooze (approved):

PR #362, opened 2021-07-22
US128833 - adding support for update-allowable-filetype action for assignments

maddiegabriel:
[US128833 - [Restrict File Ext][UI] Add new dropdown for the restrict file extension options](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fuserstory%2F600937249791)
Used by [https://github.com/BrightspaceHypermediaComponents/activities/pull/1898](https://github.com/BrightspaceHypermediaComponents/activities/pull/1898)

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -421,7 +421,6 @@ export class AssignmentEntity extends Entity {
 		return currentAllowableFileType ? [currentAllowableFileType] : [];
 	}
 
-
 	/**
 	 * @returns {object} Submission type of the assignment (including type value and type title)
 	 */
@@ -433,6 +432,9 @@ export class AssignmentEntity extends Entity {
 		return this._entity.properties.submissionType;
 	}
 
+	/**
+	 * @returns {object} Allowable filetype of the assignment (including type value and type title)
+	 */
 	allowableFileType() {
 		if (!this._entity || !this._entity.properties) {
 			return;
@@ -461,6 +463,9 @@ export class AssignmentEntity extends Entity {
 		return action.getFieldByName('submissionType').value;
 	}
 
+	/**
+	 * @returns {Array} Set of allowable filetype options for this assignment
+	 */
 	allowableFileTypeOptions() {
 		if (!this._entity) {
 			return [];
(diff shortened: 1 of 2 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -816,6 +888,13 @@ export class AssignmentEntity extends Entity {
 			fields.push({ name: 'submissionType', value: assignment.submissionType });
 		}
 
+		const shouldSaveAllowableFileType = typeof assignment.allowableFileType !== 'undefined' &&
```
jeroach, line 891: Might just be me, but I find this hard to read. Perhaps break it out into smaller variables, and then use those variables to determine the value of `shouldSaveAllowableFileType`

Would also make debugging easier :)

amauryperez (approved):

amauryperez:
LGTM

jeroach (approved):
LGTM

src/activities/assignments/AssignmentEntity.js
```diff
@@ -865,6 +943,7 @@ export class AssignmentEntity extends Entity {
 			[this.name(), assignment.name],
 			[this.instructionsEditorHtml(), assignment.instructions],
 			[this.submissionType() && String(this.submissionType().value), assignment.submissionType],
+			[this.allowableFileType() && String(this.allowableFileType().value), assignment.allowableFileType],
```
wongvincent, line 946: The `equals` function here is always returning `false` because of this line. I see it locally and on the latest quad sites.

wongvincent: Oh, is it `false` because the code in `activities` hasn't been merged yet? Could you double check locally if the dialog is still triggering with your activities changes.

maddiegabriel: That's most probably the issue - I don't see that dialog when testing locally with the changes from both repos 👍 

I'm unsure how to get the `activities` repo to pick up on this repo's changes (which it depends on) - do I need to trigger a release?

wongvincent: It should automatically be picked up by the activities CI.

PR #364, opened 2021-08-04
Mwang/us128834

cloudislife8899:


src/activities/assignments/AssignmentEntity.js
```diff
@@ -894,6 +927,11 @@ export class AssignmentEntity extends Entity {
 			fields.push({ name: 'allowableFileType', value: assignment.allowableFileType });
 		}
 
+		const allowableFileTypeCustomValue = '5';
```
wongvincent, line 930: Would be best if we didn't have to use a magic number. Although, I don't know the entity enough to suggest another way of getting the value, but if you can think of one, it would be better.

src/activities/assignments/AssignmentEntity.js
```diff
@@ -894,6 +927,11 @@ export class AssignmentEntity extends Entity {
...
+		if (assignment.allowableFileType === allowableFileTypeCustomValue) {
```
wongvincent, line 931: Should we check whether the `update-custom-allowable-file-type` action is present?
Should we check if `assignment.customAllowableFileTypes` is defined?

src/activities/assignments/AssignmentEntity.js
```diff
@@ -554,6 +563,30 @@ export class AssignmentEntity extends Entity {
 		await performSirenAction(this._token, action, fields);
 	}
 
+	/**
+	 * Sets the custom allowable filetypes of the assignment
+	 * @param {string} customAllowableFileTypes Allowable filetype option
+	 */
+	async setCustomAllowableFileTypes(customAllowableFileTypes) {
+		customAllowableFileTypes = String(customAllowableFileTypes);
+		const action = this.canEditAllowableFileType() && this._entity.getActionByName(Actions.assignments.updateCustomAllowableFileType);
```
wongvincent, line 572: should we instead be checking if it has the `update-custom-allowable-file-type` action?

src/activities/assignments/AssignmentEntity.js
```diff
@@ -554,6 +563,30 @@ export class AssignmentEntity extends Entity {
...
+		}
+
+		const fieldValue = action.getFieldByName('customAllowableFileTypes').value;
+
+		if (!fieldValue) {
+			return;
+		}
+
+		const fields = [
+			{ name: 'customAllowableFileTypes', value: customAllowableFileTypes }
```
wongvincent, line 584: Could `customAllowableFileTypes` be undefined, and if it is, do we still want to continue performing the siren action?

src/activities/assignments/AssignmentEntity.js
cloudislife8899: Added!

src/activities/assignments/AssignmentEntity.js
cloudislife8899: We need to use the id/number because first the selection needs to be unique and second because the name (text) will be translated so we cannot compare the selection against a hardcoded value

src/activities/assignments/AssignmentEntity.js
cloudislife8899: Theres a check to see if it is undefined. It does not continue if it is.

src/activities/assignments/AssignmentEntity.js
cloudislife8899: Changed!

JamieTran (approved):

src/activities/assignments/AssignmentEntity.js
```diff
@@ -554,6 +567,30 @@ export class AssignmentEntity extends Entity {
...
+	 * @param {string} customAllowableFileTypes Allowable filetype option
+	 */
+	async setCustomAllowableFileTypes(customAllowableFileTypes) {
+		customAllowableFileTypes = String(customAllowableFileTypes);
+		const action = this.canEditCustomAllowableFileTypes() && this._entity.getActionByName(Actions.assignments.updateCustomAllowableFileType);
+		if (!action) {
+			return;
+		}
+
+		const fieldValue = action.getFieldByName('customAllowableFileTypes').value;
```
jeroach, line 581: Is `action.getFieldByName('customAllowableFileTypes')` guaranteed to return a valid value?

cloudislife8899: Yes! If it is undefined I set it to an empty string.

wongvincent (approved):

PR #374, opened 2021-09-22
Mdulat/us131442 new scoring entity

mdulat:
https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fuserstory%2F608101223745&fdp=true

Create new ScoringEntity instead of tracking it within ActivityUsage Entity.

Depends on:
https://github.com/BrightspaceHypermediaComponents/activities/pull/1986
https://github.com/Brightspace/lms/pull/14361

### DIFF
--- a/src/activities/ScoringEntity.js
+++ b/src/activities/ScoringEntity.js
@@ -23,7 +23,7 @@ export class ScoringEntity extends Entity {
 	/**
 	 * @returns {bool} Whether or not the update score out of action is present
 	 */
-	 canUpdateScoring() {
+	canUpdateScoring() {
 		return this._entity && this._entity.hasActionByName(Actions.activities.scoreOutOf.update);
 	}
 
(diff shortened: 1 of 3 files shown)


src/activities/ScoringEntity.js
```diff
@@ -0,0 +1,54 @@
...
+	 */
+	scoreOutOf() {
+		return this._entity && this._entity.properties && this._entity.properties.scoreOutOf;
+	}
+
+	/**
+	 * @returns {string} Activity's grade maxPoints, if there is a grade association
+	 */
+	gradeMaxPoints() {
+		return this._entity && this._entity.properties && this._entity.properties.gradeMaxPoints;
```
mostlyfabulous, line 20: With short circuit evaluation in JS, could this return a boolean instead a string when the earlier expressions are false?

mostlyfabulous (approved):

src/activities/ScoringEntity.js
mdulat: I believe if any value is `undefined` then the whole thing will evaluate to `undefined`, and not a boolean value.

PR #376, opened 2021-09-24
fix: dirty check changes to score-out-of

mdulat:
https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fuserstory%2F608101223185&fdp=true

### DIFF
--- a/src/activities/ScoringEntity.js
+++ b/src/activities/ScoringEntity.js
@@ -27,6 +27,10 @@ export class ScoringEntity extends Entity {
 		return this._entity && this._entity.hasActionByName(Actions.activities.scoreOutOf.update);
 	}
 
+	equals(scoring) {
+		return String(this.scoreOutOf()) === scoring.scoreOutOf;
+	}
+
 	_getUpdateAction() {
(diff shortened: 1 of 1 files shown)


src/activities/ScoringEntity.js
```diff
@@ -27,6 +27,10 @@ export class ScoringEntity extends Entity {
 		return this._entity && this._entity.hasActionByName(Actions.activities.scoreOutOf.update);
 	}
 
+	equals(scoring) {
+		return String(this.scoreOutOf()) === scoring.scoreOutOf;
```
ronan-f, line 31: Is it ever possible that `scoring.scoreOutOf` might not be a string?

mdulat: Nope, it's initialized like this: `entity.scoreOutOf() ? entity.scoreOutOf().toString() : ''` and otherwise comes from an input field, so will always be a string.

ronan-f (approved):

src/activities/ScoringEntity.js
mdulat: Actually, I think I'm wrong, as the input now is a `d2l-input-number` which is a number type, not string. I'll put up another PR, thanks for asking that!

PR #382, opened 2021-10-05
US131582 - Add initial commit for `QuizSubmissionViewsEntity`

wongvincent:
Add file for `QuizSubmissionViewsEntity` and add a test file.
My story doesn't include checking for the `add` action, but I included that so that the test and entity file aren't blank.

https://rally1.rallydev.com/#/?detail=/userstory/608889244131&fdp=true

### DIFF
--- a/test/activities/quizzes/data/submissionViews/SubmissionViewsEntity.js
+++ b/test/activities/quizzes/data/submissionViews/SubmissionViewsEntity.js
@@ -1,0 +1,58 @@
+export const editableCollection = {
+	class: ['submission-views', 'collection'],
+	entities: [
+		{
+			class: ['submission-view', 'primary'],
+			rel: ['item'],
+			href:
+				'/{orgUnitId}/quizzes/{quizId}/submissionviews/7?workingCopyId=123'
+		},
+		{
+			class: ['submission-view', 'secondary'],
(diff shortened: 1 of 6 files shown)


test/activities/quizzes/data/submissionViews/SubmissionViewsEntity.js
```diff
@@ -0,0 +1,58 @@
...
+				'/{orgUnitId}/quizzes/{quizId}/submissionviews/7?workingCopyId=123'
+		},
+		{
+			class: ['submission-view', 'secondary'],
+			rel: ['item'],
+			href:
+				'/{orgUnitId}/quizzes/{quizId}/submissionviews/9?workingCopyId=123'
+		}
+	],
+	links: [
```
dustxd, line 51: Thinking back to the work I did in siren serializer...🤔  Would `actions` just not exist or show up as an empty array?

dustxd (approved):

test/activities/quizzes/data/submissionViews/SubmissionViewsEntity.js
wongvincent: Don't know, but don't think it matters.
If it breaks, then `hasActionByName` has a bug

PR #390, opened 2021-10-14
fix: dirty check

mdulat:
https://trello.com/c/lngJ1gLM/466-face-assignments-dirty-check-is-broken

### DIFF
--- a/src/activities/ScoringEntity.js
+++ b/src/activities/ScoringEntity.js
@@ -28,7 +28,7 @@ export class ScoringEntity extends Entity {
 	}
 
 	equals(scoring) {
-		return this.scoreOutOf() === scoring.scoreOutOf;
+		return this.gradeMaxPoints() === scoring.gradeMaxPoints;
 	}
 
 	_getUpdateAction() {


mdulat:
> I'm pretty ignorant to how all this grades stuff works so possibly a stupid question but do we also need to check for changes in `scoreOutOf`?

No, it's confusing, but we never change `scoreOutOf`, we only change `gradeMaxPoints`. It's because for quizzing `scoreOutOf` is something else and read-only, so we decided to track the "Grade Out Of" value that the user can update in the `gradeMaxPoints` variable only.

ronan-f:
> > I'm pretty ignorant to how all this grades stuff works so possibly a stupid question but do we also need to check for changes in `scoreOutOf`?
> 
> No, it's confusing, but we never change `scoreOutOf`, we only change `gradeMaxPoints`. It's because for quizzing `scoreOutOf` is something else and read-only, so we decided to track the "Grade Out Of" value that the user can update in the `gradeMaxPoints` variable only.

Thanks for the explanation!

PR #415, opened 2021-11-03
TA158697 (Part 2): Add mediaFileName property

coffeexcode:
## Relevant Rally
- [TA158697](https://rally1.rallydev.com/#/?detail=/task/612430665105&fdp=true): Add properties onto HM response

## Description
Add a new property for the media file name to easily support content service media files and manage files files.


brissons (approved):
Might need to add this to the `ContentMediaFileEntity` test, but LGTM!

PR #429, opened 2021-11-23
fix: DE46108 saving no score should be empty string

mdulat:
https://rally1.rallydev.com/#/?detail=/defect/615608209587&fdp=true

This fixes one small issue where saving "ungraded" did a PATCH on score-out-of endpoint with value "undefined" instead of empty string, which resulted in error.

### DIFF
--- a/src/activities/ScoringEntity.js
+++ b/src/activities/ScoringEntity.js
@@ -49,7 +49,7 @@ export class ScoringEntity extends Entity {
 		}
 
 		if (scoring.gradeMaxPoints !== this.gradeMaxPoints()) {
-			const fields = [{ name: this._getUpdateFieldName(), value: scoring.gradeMaxPoints }];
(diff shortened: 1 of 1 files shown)


mostlyfabulous (approved):

src/activities/ScoringEntity.js
```diff
@@ -49,7 +49,7 @@ export class ScoringEntity extends Entity {
 		}
 
 		if (scoring.gradeMaxPoints !== this.gradeMaxPoints()) {
-			const fields = [{ name: this._getUpdateFieldName(), value: scoring.gradeMaxPoints }];
+			const fields = [{ name: this._getUpdateFieldName(), value: scoring.gradeMaxPoints || '' }];
```
mostlyfabulous, line 52: Was `undefined` or `null` being set for no score?

mdulat: both, depending on where (it was null initially when you clicked ungraded button, and undefined if you cleared the input manually).

PR #433, opened 2021-11-25
Changed URL to tool specific

ronan-f:
I was thinking of changing this to `/activities` as categories is now used in multiple tools, however, as Margaret pointed out that doesn't represent a real URL in our system and the actual use of PATCHing categories will be done against a specific tools URL.


mdulat (approved):
Thanks for that.

PR #437, opened 2021-12-08
Added tests and refactored assignments save to avoid depending on AF specific PUT action

ronan-f:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fdefect%2F618752964661

Apologies for the oversized PR. The root cause of this permissions check is that Assignments was implicitly depending on the `PUT` action which was designed for AF.

[We were overriding the action's method](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/assignments/AssignmentEntity.js#L896) inside `save()` which worked provided the `PUT` action was present on the assignment entity itself. 

If the `PUT` action wasn't present then [this check](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/d9b3c85d7549997b3475fe2f6f84b1d5e6440e31/src/activities/assignments/AssignmentEntity.js#L885) failed and the save became a no-op.

The `PUT` action gets included provided a user has the `ManageDelivery` permission: 

https://github.com/Brightspace/lms/blob/master/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs#L530

https://github.com/Brightspace/lms/blob/master/le/dropbox/D2L.LE.Dropbox/Utilities/Access/DropboxAccessChecker.cs#L103

This fix ensures that the `save` method instead checks each update action individually and formats them into a single request.

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -1,6 +1,6 @@
 import { Entity } from '../../es6/Entity';
 import { Actions, Rels, Classes } from '../../hypermedia-constants';
-import { performSirenAction } from '../../es6/SirenAction';
+import { performSirenAction, performSirenActions } from '../../es6/SirenAction';
 
 const actions = {
 	delete: 'delete-folder',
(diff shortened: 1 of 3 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -345,75 +349,6 @@ export class AssignmentEntity extends Entity {
 		return organizationHref && organizationHref.href;
 	}
 
-	/**
-	 * Sets the assignment type to group using a default group category
-	 */
-	async setToGroupAssignmentType() {
```
ronan-f, line 351: I don't think this or the 2 methods below are being used anywhere

mdulat (approved):
Nice

PR #438, opened 2021-12-16
DE46154 Add Grade Category `maxPoints` and `autoPoints` properties & Grade Candidate `autoPoints` property

mostlyfabulous:
[DE46153](https://rally1.rallydev.com/#/?detail=/defect/616899740963&fdp=true): FACE Assignment > When choose to link to existing grade item as part of grade category with distributed points, grade out of should not be editable
[DE46154](https://rally1.rallydev.com/#/?detail=/defect/616901920397&fdp=true): FACE Assignment > Updating Grade Out Of value on an assignment linked to a Grade item under grade category with distributed points, doesn't respect the distribution

https://github.com/BrightspaceHypermediaComponents/activities/pull/2248

### DIFF
--- a/src/activities/GradeCandidateEntity.js
+++ b/src/activities/GradeCandidateEntity.js
@@ -25,6 +25,15 @@ export class GradeCandidateEntity extends Entity {
 			return entity.getLinkByRel(Rels.Grades.grade).href;
 		}
 	}
+	/**
+	 * @returns {bool} True if grade-candidate is `autoPoints` i.e. maxPoints are set automatically
+	 */
+	isAutoPoints() {
+		const associateGradeAction = this._entity && this._entity.getActionByName(Actions.activities.associateGrade.associateGrade);
+		if (!associateGradeAction) return false;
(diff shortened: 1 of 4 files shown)


src/activities/GradeCandidateEntity.js
```diff
@@ -25,6 +25,15 @@ export class GradeCandidateEntity extends Entity {
 			return entity.getLinkByRel(Rels.Grades.grade).href;
 		}
 	}
+	/**
+	 * @returns {bool} True if grade-candidate is `autoPoints` i.e. maxPoints are set automatically
+	 */
+	isAutoPoints() {
```
mdulat, line 31: I think this should go into the `GradeEntity` instead of here, so that it can be retrieved from the grade's autoPoints property, instead of from an action field. This action shouldn't have an autoPoints field.

mostlyfabulous: Was just about to mention the link between this and https://github.com/Brightspace/lms/pull/17217#discussion_r770940680

I'll try and rework things 😃

mdulat (approved):

PR #449, opened 2022-01-27
Create new end of sequence entity

ianlaidlaw:
Creates a new representation for the Sequence's End of Sequence Entity. Right now the only thing we need to get from it is the properties from the summary subentity. More stuff can be added to it as it is needed.

https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F622544965323&fdp=true

related sequences pr: https://github.com/BrightspaceHypermediaComponents/sequences/pull/382

### DIFF
--- a/src/sequences/EndOfSequenceEntity.js
+++ b/src/sequences/EndOfSequenceEntity.js
@@ -1,0 +1,12 @@
+import { Entity } from '../es6/Entity';
+
+export class EndOfSequenceActivityEntity extends Entity {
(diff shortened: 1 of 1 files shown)


ChrisKraljevicD2L (approved):

brissons (commented):
I would maybe add a method comment with `@returns` etc, but lgtm!

brissons (approved):

PR #455, opened 2022-02-15
US136045: Add types to weblinks

coffeexcode:
## Relevant Rally
- [US136045](https://rally1.rallydev.com/#/?detail=/userstory/626774156023&fdp=true): Add Banner with name of file

## Description
This PR adds support in siren to easily get the type of weblink (google drive, one drive, or default)

### DIFF
--- a/src/activities/content/ContentWebLinkEntity.js
+++ b/src/activities/content/ContentWebLinkEntity.js
@@ -67,6 +67,9 @@ export class ContentWebLinkEntity extends ContentWorkingCopyEntity {
 		return this._entity && this._entity.properties && this._entity.properties.url;
 	}
 
+	/**
+	 * @returns {string} The type of weblink
+	 */
 	type() {
 		if (this._entity) {
 			if (this._entity.hasClass(WebLinkTypes.GoogleDrive)) {
--- a/src/activities/content/ContentWebLinkEntity.js
+++ b/src/activities/content/ContentWebLinkEntity.js
@@ -72,11 +72,11 @@ export class ContentWebLinkEntity extends ContentWorkingCopyEntity {
 			if (this._entity.hasClass(WebLinkTypes.GoogleDrive)) {
 				return WebLinkTypes.GoogleDrive;
 			} else if (this._entity.hasClass(WebLinkTypes.OneDrive)) {
-				return WebLinkTypes.OneDrive
+				return WebLinkTypes.OneDrive;
 			}
 		}
 
-		return 'default';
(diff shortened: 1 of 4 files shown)


src/activities/content/ContentWebLinkEntity.js
```diff
@@ -61,6 +61,18 @@ export class ContentWebLinkEntity extends ContentWorkingCopyEntity {
 		return this._entity && this._entity.properties && this._entity.properties.url;
 	}
 
+	type() {
```
brissons, line 64: nit: up to you, but I would rename this to getWebLinkType()

src/activities/content/ContentWebLinkEntity.js
```diff
@@ -61,6 +61,18 @@ export class ContentWebLinkEntity extends ContentWorkingCopyEntity {
...
+		if (this._entity) {
+			if (this._entity.hasClass(Classes.webLink.googleDriveLink)) {
+				return Classes.webLink.googleDriveLink;
+			} else if (this._entity.hasClass(Classes.webLink.oneDriveLink)) {
+				return Classes.webLink.oneDriveLink;
+			}
+		}
+
+		return 'default';
```
brissons, line 73: I wonder if we should instead move [this](https://github.com/BrightspaceHypermediaComponents/activities/pull/2385/files#diff-af9a9f9b02aff01e8acd70627d7cfecb271c14f214802866805ccf794acce65dR1) and [this](https://github.com/BrightspaceHypermediaComponents/activities/pull/2386/files#diff-af9a9f9b02aff01e8acd70627d7cfecb271c14f214802866805ccf794acce65dR8) into `ContentWebLinkEntity` (similar to files). That way we don't have multiple lists keeping track of the same thing and having to match them; we can just consolidate it here.

We probably also just need one list rather than separating them into WebLinkTypes and CloudDriveLinks.

src/activities/content/ContentWebLinkEntity.js
coffeexcode: Normally I would agree, but I feel `type()` fits in better with what is currently here (most of these entities have getters that are just `propName()`

prajapac: Do we wanna add a comment about the return context as we do for other properties?

prajapac (approved):
LGTM +/- Scott's nit, both seem reasonable to me. Also, comment about return type comment.

prajapac (approved):

PR #456, opened 2022-02-17
US136096 - adding ActivityEvaluationStatusEntity and tests

quinnbudan:
[US136096](https://rally1.rallydev.com/#/?detail=/userstory/626871691749&fdp=true): [UCE] existing completion summary api data source

Adding the `ActivityEvaluationStatusEntity` to be used in the activities repo for the new completion summary component.

### DIFF
--- a/src/activities/ActivityEvaluationStatusEntity.js
+++ b/src/activities/ActivityEvaluationStatusEntity.js
@@ -1,0 +1,90 @@
+import { Entity } from '../es6/Entity.js';
+import { Rels } from '../hypermedia-constants';
+
+const supportedActivityTypes = [
+	'assignment',
+	'topic',
+	'quiz',
+];
+
+export class ActivityEvaluationStatusEntity extends Entity {
+	/**
+	 * @return {string|null} The activity type or null if the entity doesn't not have an expected class.
+	 */
+	getActivityType() {
+		if (this._entity) {
+			for (const hmClass of supportedActivityTypes) {
+				if (this._entity.hasClass(hmClass)) {
+					return hmClass;
+				}
+			}
+		}
+		return null;
+	}
+
+	/**
+	 * @returns {int|null} The number of assignments of this activity
+	 */
+	numAssigned() {
+		return this._entity ? Number(this._entity.properties.assigned) : null;
+	}
+
+	/**
+	 * @returns {int|null} The number of completions of this activity
+	 */
+	numCompleted() {
+		return this._entity ? Number(this._entity.properties.completed) : null;
+	}
+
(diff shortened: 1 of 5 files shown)


louiscoste (approved):

src/activities/ActivityEvaluationStatusEntity.js
```diff
@@ -0,0 +1,90 @@
...
+	numResubmissions() {
+		return this._entity ? Number(this._entity.properties.resubmissions) : null;
+	}
+
+	/**
+	 * @returns {string|null} The path to the assessment application, for all assessments
+	 */
+	assesAllApplication() {
+		const entity = this._entity.getSubEntityByRel(Rels.Assessments.assessAllApplication);
+		return entity ? entity.properties.path : null;
```
TomMichielsD2L, line 72: Nit: I am starting to doubt we have optional chaining but `return entity?.properties.path`

quinnbudan: We don't have optional chaining in this repo, it depends on polymer and is also depended on by activities!

src/activities/ActivityEvaluationStatusEntity.js
```diff
@@ -0,0 +1,90 @@
...
+			}
+		}
+		return null;
+	}
+
+	/**
+	 * @returns {int|null} The number of assignments of this activity
+	 */
+	numAssigned() {
+		return this._entity ? Number(this._entity.properties.assigned) : null;
```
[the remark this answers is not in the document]
TomMichielsD2L: I think I understand that this is supposed to be a class to wrap the basic entity. Is that accurate? If yes, should this have a constructor that just returns a null object if the activity type is not supported?

src/activities/ActivityEvaluationStatusEntity.js
TomMichielsD2L: :(

src/activities/ActivityEvaluationStatusEntity.js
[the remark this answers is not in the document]
quinnbudan: The reason I returned null was to be consistent with some of the entities I looked at. There's no real use case yet since the component that'll use this is still WIP, but basically the component has to handle these things being null.

I definitely like the idea of defaulting to 0 if we don't have an entity! One thing that's a bit weird about it though is that if we do this we're almost lying about the situation, because the entity isn't actually loaded and we're saying "the entity says there are 0 assignments of this activity"

TomMichielsD2L (approved):

PR #482, opened 2022-05-19
Fix shouldUpdate in entity-mixin-lit

svanherk:
I'm a bit worried about this change, but in theory it should only be a good thing.  Basically, anything using both the `EntityMixinLit` and `core`'s `LocalizeDynamicMixin` isn't actually waiting until the langterms are loaded before rendering.  This is probably only actually noticeable in the OSLO case, where langterms take a while to return.

I think the fix below is the correct way to do this - `EntityMixinLit` gets the chance to say "no, don't update, we don't have an `href`/`token`" but will then default to other mixin's requirements if its good to go with the update.

Thoughts?

### DIFF
--- a/src/mixin/entity-mixin-lit.js
+++ b/src/mixin/entity-mixin-lit.js
@@ -35,7 +35,7 @@ export const EntityMixinLit = superclass => class extends superclass {
 			this.href && this.token) {
 			this._getEntity();
 		}
-		return this.href && this.token;
+		return this.href && this.token ? super.shouldUpdate(changedProperties) : false;
 	}
 
 	_entityHasChanged(newValue, oldValue) {


dlockhart (approved):
I think this is the right change -- it may break some tests in other repos potentially since rendering may happen a bit later.

src/mixin/entity-mixin-lit.js
```diff
@@ -35,7 +35,7 @@ export const EntityMixinLit = superclass => class extends superclass {
 			this.href && this.token) {
 			this._getEntity();
 		}
-		return this.href && this.token;
+		return this.href && this.token ? super.shouldUpdate(changedProperties) : false;
```
dlockhart, line 38: Actually... `localize-mixin.js` had to do a bunch of extra work to accomplish something similar, so I'm worried something is getting lost here. Like it's remembering the list of updated properties when it returns `false` so that later when it can call `super.shouldUpdate(changedProperties)` it can include all the ones that it skipped.

svanherk: Oh interesting, I can look into that

dlockhart: It's possible that it's overkill... you could probably just remember if `super.shouldUpdate(changedProperties)` would have ever returned `true` (when you're returning `false` because `href`/`token` aren't defined yet) and then if so, return that value.

svanherk: Yeah interesting, I think it's overkill for the use cases I know about because the `localizeMixin` doesn't actually _use_ the `changedProperties`.  But there could be another mixin after it that does... could this be a controller mixins could use? 🤔

PR #484, opened 2022-05-25
test: Migrate tests off of WCT - Part 1

svanherk:
The tests are doing some hardcoding of how to get the `siren-parser` stuff, which breaks when we update the name for `npm`.  Instead of just adjusting the path I'm just going to migrate all the tests, which I've wanted to do for a while anyways.  This is PR 1 of 3 (and then I'll probably do a linting one).


test/alignments/AlignmentsCollectionEntity.test.js
```diff
@@ -1,17 +1,19 @@
 import { AlignmentsCollectionEntity } from '../../src/alignments/AlignmentsCollectionEntity.js';
```
svanherk, line 1: This test wasn't running before, so you'll see a bunch of fixes below to get it working.

test/index.html
```diff
@@ -13,21 +13,6 @@
 		/* global WCT */
 		'use strict';
 		WCT.loadSuites([
```
svanherk, line 15: There's a bunch of tests missing from here that weren't running before.  In this PR, that's `mixin/simple-entity-mixin-test.html`, `AlignmentsCollectionEntity.html`, and `FilesHomeEntity.html`.

dlockhart (approved):
Thanks for going the extra mile on these to get them into better shape!

PR #499, opened 2022-07-08
US140433 Add discussion topic entity with name field related functions

dustxd:
### Summary of Changes
* Added Discussion Topic Entity
* Added functions for fetching and setting name of topic, as well as permission check
* Added tests for the name/title field (Note that the only difference between the two test data files is the presence of `actions`)

### Rally Story
[US140433](https://rally1.rallydev.com/#/?detail=/userstory/640968148381&fdp=true): [ui] Edit topic title field

### DIFF
--- a/src/activities/discussions/DiscussionTopicEntity.js
+++ b/src/activities/discussions/DiscussionTopicEntity.js
@@ -1,0 +1,9 @@
+import { Actions } from '../../hypermedia-constants.js';
+import { Entity } from '../../es6/Entity.js';
+import { performSirenAction } from '../../es6/SirenAction.js';
+
+/**
+ * DiscussionTopicEntity class representation of a D2L Discussion Topic.
+ */
+export class DiscussionTopicEntity extends Entity {
+}
--- a/src/activities/discussions/DiscussionTopicEntity.js
+++ b/src/activities/discussions/DiscussionTopicEntity.js
@@ -19,4 +19,26 @@ export class DiscussionTopicEntity extends Entity {
 	canEditName() {
 		return this._entity && this._entity.hasActionByName(Actions.discussions.topic.updateName);
 	}
+
+	/**
+	 * @summary Formats action and fields if topic name has changed and user has edit permission
+	 * @param {object} topic the topic that's being modified
+	 * @returns {object} the appropriate action/fields to update
+	 */
+	_formatUpdateNameAction(topic) {
+		const { name } = topic || {};
+
+		if (!name) return;
+		if (!this._hasFieldValueChanged(name, this.name())) return;
+		if (!this.canEditName()) return;
+
+		const action = this._entity.getActionByName(Actions.discussions.topic.updateName);
+		const fields = [{ name: 'name', value: name }];
+
+		return { action, fields };
+	}
+
+	_hasFieldValueChanged(currentValue, initialValue) {
+		return currentValue !== initialValue;
+	}
 }
--- a/src/activities/discussions/DiscussionTopicEntity.js
+++ b/src/activities/discussions/DiscussionTopicEntity.js
@@ -1,6 +1,6 @@
 import { Actions } from '../../hypermedia-constants.js';
 import { Entity } from '../../es6/Entity.js';
-import { performSirenAction } from '../../es6/SirenAction.js';
+import { performSirenActions } from '../../es6/SirenAction.js';
 
 /**
  * DiscussionTopicEntity class representation of a D2L Discussion Topic.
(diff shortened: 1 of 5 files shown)


mostlyfabulous:
We will probably need the `_formatUpdateNameAction` like in the AssignmentEntity & the [QuizEntity](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/ee456e4f37886126cd611da175eeb50d95fb5273/src/activities/quizzes/QuizEntity.js#L693) 
```
	/**
	 * Checks if quiz name has changed and if so returns the appropriate action/fields to update
	 * @param {object} quiz the quiz that's being modified
	 */

	_formatUpdateNameAction(quiz) {
		const { name } = quiz || {};

		if (!name) return;
		if (!this._hasNameChanged(name)) return;

		const action = this._entity.getActionByName(Actions.quizzes.updateName);

		if (!action) {
			return;
		}

		const fields = [
			{ name: 'name', value: name },
		];

		return { action, fields };
	}
```
For when we call save on the topic later like in quiz:
````
	async save(quiz) {
		if (!quiz) return;
		const updateNameAction = this.canEditName() ? this._formatUpdateNameAction(quiz) : null;
```

dustxd:
> We will probably need the `_formatUpdateNameAction` like in the AssignmentEntity & the [QuizEntity](https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/ee456e4f37886126cd611da175eeb50d95fb5273/src/activities/quizzes/QuizEntity.js#L693)
> 
> ```
> 	/**
> 	 * Checks if quiz name has changed and if so returns the appropriate action/fields to update
> 	 * @param {object} quiz the quiz that's being modified
> 	 */
> 
> 	_formatUpdateNameAction(quiz) {
> 		const { name } = quiz || {};
> 
> 		if (!name) return;
> 		if (!this._hasNameChanged(name)) return;
> 
> 		const action = this._entity.getActionByName(Actions.quizzes.updateName);
> 
> 		if (!action) {
> 			return;
> 		}
> 
> 		const fields = [
> 			{ name: 'name', value: name },
> 		];
> 
> 		return { action, fields };
> 	}
> ```
> 
> For when we call save on the topic later like in quiz:
> 
> ```
> 	async save(quiz) {
> 		if (!quiz) return;
> 		const updateNameAction = this.canEditName() ? this._formatUpdateNameAction(quiz) : null;
> ```
> ```

Now that you bring this up - I think we don't need the individual `setName` action at all. I can't think of an instance where we don't want the name field to be part of the group checks and updates but maybe Margaret can comment on this. I might as well just add the `save` and `equals` functions too.

mdulat:
> Now that you bring this up - I think we don't need the individual `setName` action at all. I can't think of an instance where we don't want the name field to be part of the group checks and updates but maybe Margaret can comment on this. I might as well just add the `save` and `equals` functions too.

Agreed. No need for `setName` because we're not doing auto-save. So `_formatUpdateNameAction`, `save` and `equals` instead makes sense.

src/activities/discussions/DiscussionTopicEntity.js
```diff
@@ -21,16 +21,58 @@ export class DiscussionTopicEntity extends Entity {
...
+			if (current !== initial) {
+				return false;
+			}
 		}
 
+		return true;
+	}
+
+	/**
+	 * @summary Fires all the formatted siren actions collectively
```
mostlyfabulous, line 42: Love this.

src/activities/discussions/DiscussionTopicEntity.js
```diff
@@ -21,16 +21,58 @@ export class DiscussionTopicEntity extends Entity {
...
+
+		await performSirenActions(this._token, sirenActions);
+	}
+
+	/**
+	 * @summary Formats action and fields if topic name has changed and user has edit permission
+	 * @param {object} topic the topic that's being modified
+	 * @returns {object} the appropriate action/fields to update
+	 */
+	_formatUpdateNameAction(topic) {
```
mostlyfabulous, line 62: I found it hard to read the quiz entity when the action was at the bottom of the file, away from the property (name) getter. Maybe we can follow the style of the assignment entity which kept them close?
https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/assignments/AssignmentEntity.js#L17-L32

mostlyfabulous (approved):

mdulat (approved):

src/activities/discussions/DiscussionTopicEntity.js
dustxd: Thanks 😁 I renamed and tweaked a few things so it looks different from QuizEntity. I was worried that would be confusing!

PR #505, opened 2022-07-27
fix: Updates to allow load more paging

zwilson2:
### Rally Story:
* https://rally1.rallydev.com/#/255540250444d/backlog?detail=%2Fuserstory%2F642767486081

### Related PRs:
* https://github.com/Brightspace/ipsis-sis-course-merge/pull/17

Looking for initial feedback as I'm not sure I like this change as is.

### DIFF
--- a/src/ipsis/course-merge/CourseMergeOfferingCollectionEntity.js
+++ b/src/ipsis/course-merge/CourseMergeOfferingCollectionEntity.js
@@ -12,6 +12,10 @@ export class CourseMergeOfferingCollectionEntity extends Entity {
 		return this._entity.entities;
 	}
 
+	prependCourseMergeOfferings(previousCourseMergeOfferingCollectionEntity) {
+		this._entity.entities.unshift(...previousCourseMergeOfferingCollectionEntity._entity.entities);
+	}
+
 	userOwnedByMultipleSourceSystems() {
 		return this._entity?.properties?.userOwnedByMultipleSourceSystems;
 	}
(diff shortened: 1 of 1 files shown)


src/ipsis/course-merge/CourseMergeOfferingCollectionEntity.js
```diff
@@ -3,14 +3,13 @@
...
 import { Entity } from '../../es6/Entity.js';
-import { Rels } from '../../hypermedia-constants.js';
 
 export class CourseMergeOfferingCollectionEntity extends Entity {
 	courseMergeOfferings() {
 		if (!this._entity) {
 			return;
 		}
-		return this._entity.getSubEntitiesByRel(Rels.IPSIS.SISCourseMerge.courseMergeOffering);
+		return this._entity.entities;
```
zwilson2, line 12: This change is needed due to how the `siren-parser` `Entity` (`_entity`) works. As part of it's constructor, it parses `entities` into objects like `entitiesByRel` for easier lookup later. Because of this, updating `entities` doesn't update `entitiesByRel` which means `getSubEntitiesByRel` will still only return for the current page instead of `entities`. And the only time we call the `Entity` constructor is when we receive the new `href`.

JeremyWinkler: I see why we wouldn't like the change, but we should never be returning anything else in the subentities so I feel like this would've been ok from the start.

JeremyWinkler (approved):

JeremyWinkler (approved):

PR #506, opened 2022-08-02
Update pageSize to indicate if less than normal page size

zwilson2:
### Rally Story:
* [US141251](https://rally1.rallydev.com/#/?detail=/userstory/642767486081&fdp=true): Course Merge - UI - Paging

### Predecessor PRs:
* https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/505

### Related PRs:
* https://github.com/Brightspace/ipsis-sis-course-merge/pull/18

This fixes the issue where we'll have loaded 40 of 50 items but it will still say `Load 20 More`. See thread: https://d2l.slack.com/archives/C0PHG3QB0/p1659451002535809?thread_ts=1659125662.045049&cid=C0PHG3QB0

### DIFF
--- a/src/ipsis/course-merge/CourseMergeOfferingCollectionEntity.js
+++ b/src/ipsis/course-merge/CourseMergeOfferingCollectionEntity.js
@@ -33,9 +33,14 @@ export class CourseMergeOfferingCollectionEntity extends Entity {
 	}
 
 	pageSize() {
+		return this._pagingInfo()?.pageSize;
+	}
+
+	loadMorePageSize() {
 		const pageSize = this._pagingInfo()?.pageSize;
 		const totalCount = this.totalCount() ?? 0;
 		const courseMergeOfferingsLength = this.courseMergeOfferings()?.length ?? 0;
+		// if pageSize is larger than the number remaining items, return the number of remaining items to be loaded
 		if (totalCount < courseMergeOfferingsLength + (pageSize ?? 0)) {
 			return totalCount - courseMergeOfferingsLength;
 		}
--- a/src/ipsis/course-merge/CourseMergeOfferingCollectionEntity.js
+++ b/src/ipsis/course-merge/CourseMergeOfferingCollectionEntity.js
@@ -33,7 +33,13 @@ export class CourseMergeOfferingCollectionEntity extends Entity {
 	}
 
 	pageSize() {
-		return this._pagingInfo()?.pageSize;
+		const pageSize = this._pagingInfo()?.pageSize;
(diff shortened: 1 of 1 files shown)


src/ipsis/course-merge/CourseMergeOfferingCollectionEntity.js
```diff
@@ -33,7 +33,13 @@ export class CourseMergeOfferingCollectionEntity extends Entity {
...
 
 	pageSize() {
-		return this._pagingInfo()?.pageSize;
+		const pageSize = this._pagingInfo()?.pageSize;
+		const totalCount = this.totalCount() ?? 0;
+		const courseMergeOfferingsLength = this.courseMergeOfferings()?.length ?? 0;
+		if (totalCount < courseMergeOfferingsLength + (pageSize ?? 0)) {
+			return totalCount - courseMergeOfferingsLength;
+		}
+		return pageSize;
```
JeremyWinkler, line 42: Maybe I'm misunderstanding what this is supposed to do, but I'm assuming we want to return the length of the current page?
With the following:
```
pageSize = 20
totalCount = 5
couseMergeOfferingsLength = 5
```
Then `totalCount < courseMergeOfferingsLength + (pageSize ?? 0)` would be `true` so we'd return `0`.
Should this not just be returning `courseMergeOfferingsLength`?

zwilson2: We want to return the length of the remaining items. Without this change, as mentioned in the description of the PR, if you have 40 of 50 items loaded, the `pageSize` will still be 20 and the Load More component will say `Load 20 More` even though the 40 of 50 items suggests that there are only 10 left.
To the "Then totalCount < courseMergeOfferingsLength + (pageSize ?? 0) would be true so we'd return 0.", returning 0 is okay. If there are no more items, we shouldn't have gotten a `next` href and the load more button shouldn't show up anymore.

I can add a comment to describe the goal of this.

zwilson2: Actually, I'm going to move this to a separate function as we should have a way to return just the `pageSize`.

JeremyWinkler (approved):

src/ipsis/course-merge/CourseMergeOfferingCollectionEntity.js
JeremyWinkler: Yea I thought it was returning current page size, but you wanted next page size which is what confused me.

PR #517, opened 2022-08-18
US142118 remove references to RichtextEditorConfig

mdulat:
https://rally1.rallydev.com/#/?detail=/userstory/644280021699&fdp=true

### DIFF
--- a/docs/AssignmentEntity.html
+++ b/docs/AssignmentEntity.html
@@ -19,7 +19,7 @@
 
     <h1 class="page-title">Class: AssignmentEntity</h1>
 
-    
+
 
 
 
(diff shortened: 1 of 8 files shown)


docs/AssignmentEntity.html
```diff
@@ -2568,29 +2568,28 @@ <h5>Returns:</h5>
...
 
 
 
-        
-            
 
-    
 
-    
-    <h4 class="name" id="instructionsRichTextEditorConfig"><span class="type-signature"></span>instructionsRichTextEditorConfig<span class="signature">()</span><span class="type-signature"> &rarr; {object}</span></h4>
```
dustxd, line 2590: rip all these huge diffs...most are spacing changes only, seems like this is the only change in this file?

dustxd (approved):

dustxd:
A quick search in `siren-sdk` seems to indicate that content also has rich text, but I guess it's not quite the same...?

martinwang-d2l (approved):

PR #521, opened 2022-09-02
feat: US140677 Add topic delete entities and actions

wkerrd2l:
[US140677](https://rally1.rallydev.com/#/?detail=/userstory/641654621783&fdp=true)

### DIFF
--- a/test/activities/discussions/DiscussionTopicEntity.test.js
+++ b/test/activities/discussions/DiscussionTopicEntity.test.js
@@ -75,6 +75,27 @@ describe('DiscussionTopicEntity', () => {
 		});
 	});
 
+	describe('delete', () => {
+		it('delete Topic', async() => {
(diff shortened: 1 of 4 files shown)


dustxd:
Can you add some tests for delete? I think one test making sure that the delete action is actually getting called and one test attempting to call delete without permissions (i.e. when the action is missing)

mostlyfabulous (approved):
Don't forget to add 'US140677' or whatever the Rally ID is in commit messages.

test/activities/discussions/DiscussionTopicEntity.test.js
```diff
@@ -75,6 +75,27 @@ describe('DiscussionTopicEntity', () => {
...
+			const discussionTopic = new DiscussionTopicEntity(editableEntity);
+
+			await discussionTopic.delete();
+			expect(fetchMock.called()).to.be.true;
+		});
+
+		it('cannot delete topic', async() => {
+			const discussionTopic = new DiscussionTopicEntity(nonEditableEntity);
+			await discussionTopic.delete();
+			expect(fetchMock.done());
```
dustxd, line 91:
```suggestion
			expect(fetchMock.called()).to.be.false;
```

PR #536, opened 2022-10-18
Functions for handling new timing editor & timer settings dialog

mxg895:
https://rally1.rallydev.com/#/16349626795d/custom/260792024672?detail=%2Fuserstory%2F643174099233
- Renamed some functions previously created for this story
- Added functions for combining logics for the editor & dialog
- Test cases for timing entities containing the new subentity.

### DIFF
--- a/src/activities/quizzes/timing/QuizTimingEntity.js
+++ b/src/activities/quizzes/timing/QuizTimingEntity.js
@@ -38,8 +38,17 @@ export class QuizTimingEntity extends Entity {
 		return entity && entity.hasActionByName(Actions.quizzes.timing.updateTimeLimit);
 	}
 
+	canEditTimingTimeLimit() {
+		const entity = this.getTimerSettingsSubEntity();
+		return entity && entity.hasActionByName(Actions.quizzes.timing.updateTimeLimit);
+	}
+
 	canEditTimeLimitType() {
-		return this.getTimeLimitTypesSubEntity()?.hasActionByName(Actions.quizzes.timing.updateTimeLimitType);
+		return this.getTimerSettingsSubEntity()?.hasActionByName(Actions.quizzes.timing.updateTimeLimitType);
+	}
+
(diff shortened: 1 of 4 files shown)


src/activities/quizzes/timing/QuizTimingEntity.js
```diff
@@ -38,8 +38,17 @@ export class QuizTimingEntity extends Entity {
 		return entity && entity.hasActionByName(Actions.quizzes.timing.updateTimeLimit);
 	}
 
+	canEditTimingTimeLimit() {
+		const entity = this.getTimerSettingsSubEntity();
+		return !!entity?.hasActionByName(Actions.quizzes.timing.updateTimeLimit);
```
dustxd, line 43: can we follow the syntax above instead of using double bang?

dustxd (approved):

src/activities/quizzes/timing/QuizTimingEntity.js
mxg895: I just tried using the above syntax. It seems that this breaks the test case where I want to test that entities with the flag off structure should expect this to be false. Without the double bang, this will just return undefined (I guess the difference with this one is that `entity` just doesn't exist in flag off structure, but for flag off functions on can edit x, a read-only version of recommended & enforced subentities still exist in the structure. I could try to expect change this (and other blocks with `!!`) to something like:
```
if (!entity) return false;
return entity.hasActionByName(Actions.quizzes.timing.updateTimeLimit);
```
if we don't want double bang?

austinkwok1997 (approved):

src/activities/quizzes/timing/QuizTimingEntity.js
dustxd: I think in general we just do `if (!entity) return;` if we expect entity to be undefined

It's fine~ You could just keep it as is 😃

PR #541, opened 2022-10-25
US142821 Add action to delete previous category and associate to a new category

mostlyfabulous:
[US142821](https://rally1.rallydev.com/#/?detail=/userstory/645735331867&fdp=true): [api] Create choose forum dialog

This let's the categories entity parse the new action serialized from here: https://github.com/Brightspace/lms/pull/29199

### DIFF
--- a/src/activities/CategoriesEntity.js
+++ b/src/activities/CategoriesEntity.js
@@ -122,14 +122,14 @@ export class CategoriesEntity extends Entity {
 		}
 	}
 
-	_generateDeletePreviousAndAssociateWithNewCategory(category, shouldDeleteDefaultCategory) {
+	_generateDeletePreviousAndAssociateWithNewCategory(category, shouldDeletePreviousCategory) {
 		const categoryEntity = this._getCategoryById(category.categoryId);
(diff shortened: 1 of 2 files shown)


src/activities/CategoriesEntity.js
```diff
@@ -122,18 +122,33 @@ export class CategoriesEntity extends Entity {
...
+
 	_hasCategoryIdChanged(categoryId) {
 		const selectedCategory = this.getSelectedCategory();
 		const initialId = selectedCategory && selectedCategory.properties.categoryId;
 
 		return categoryId !== initialId;
 	}
 
-	async save(category) {
+	async save(category, shouldDeleteDefaultCategory = false) {
```
mostlyfabulous, line 145: Will update usages of the save to take the extra param, but setting it `false` here for safety.

dustxd (approved):

src/activities/CategoriesEntity.js
dustxd: You mean we’re only updating the calls in discussion right? With the default parameter it shouldn’t affect existing usages

PR #542, opened 2022-10-26
US142491 getSelectedCategoryHref

mostlyfabulous:
[US142491](https://rally1.rallydev.com/#/?detail=/userstory/644998737805&fdp=true): [ui] Create change forum dialog

Supports: https://github.com/BrightspaceHypermediaComponents/activities/pull/3151

### DIFF
--- a/src/activities/CategoriesEntity.js
+++ b/src/activities/CategoriesEntity.js
@@ -89,6 +89,29 @@ export class CategoriesEntity extends Entity {
 		return Number(category.categoryId) === Number(selectedCategory.properties.categoryId);
 	}
 
+	getSelectedCategoryHref(categoryId) {
+		if (!this._entity) {
+			return;
+		}
+
+		const subEntities = this._entity.getSubEntitiesByClass(Classes.activities.category);
+
+		let selectedCategory;
+		if (subEntities) {
+			selectedCategory = subEntities.find(category => Number(category.properties.categoryId) === categoryId);
+		}
+
+		if (!selectedCategory) return;
+
+		let action = selectedCategory.getActionByName(Actions.activities.categories.deselect);
+		if (!action) action = selectedCategory.getActionByName(Actions.activities.categories.select);
+		if (!action) return;
+
+		const { href } = action;
+
+		return href;
+	}
+
 	_generateDeselectCategoryAction() {
 		if (!this.canEditCategories()) return;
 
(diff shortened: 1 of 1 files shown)


dustxd (approved):

src/activities/CategoriesEntity.js
```diff
@@ -89,6 +89,29 @@ export class CategoriesEntity extends Entity {
...
+		const subEntities = this._entity.getSubEntitiesByClass(Classes.activities.category);
+
+		let selectedCategory;
+		if (subEntities) {
+			selectedCategory = subEntities.find(category => Number(category.properties.categoryId) === categoryId);
+		}
+
+		if (!selectedCategory) return;
+
+		let action = selectedCategory.getActionByName(Actions.activities.categories.deselect);
```
mostlyfabulous, line 106: If the user chooses the same category, this let's us get the href and set it for consistency.

dustxd: Like one that wasn’t marked as selected? I thought it would always be deselect…?

src/activities/CategoriesEntity.js
```diff
@@ -89,6 +89,29 @@ export class CategoriesEntity extends Entity {
...
+	getSelectedCategoryHref(categoryId) {
+		if (!this._entity) {
+			return;
+		}
+
+		const subEntities = this._entity.getSubEntitiesByClass(Classes.activities.category);
+
+		let selectedCategory;
+		if (subEntities) {
+			selectedCategory = subEntities.find(category => Number(category.properties.categoryId) === categoryId);
```
dustxd, line 101: TODO if we have time later - do the number check here. I know you handled it in activities so this works but my bad for not maintaining consistency here 😅

src/activities/CategoriesEntity.js
mostlyfabulous: `deselect` only appears once in the collection of categories (it indicates that category is the currently selected category at serialization when the GET is made, but not at present in the UI).

image

src/activities/CategoriesEntity.js
mostlyfabulous: Added the cast

src/activities/CategoriesEntity.js
dustxd: The way we use it right now (in activities) persists the categories collection and refetches the collection so technically the selected category should always be updated I think, but of course it's always safer to do this! Just curious if you actually ran into issues because of this 👀

PR #544, opened 2022-10-26
US144752: Participation Options

martinwang-d2l:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F665853120201

SirenSDK support for participation options. 

Radio options, with three choices: 
- defaultParticipation 
- anonymous
- startThreadToView

### DIFF
--- a/src/activities/discussions/DiscussionTopicEntity.js
+++ b/src/activities/discussions/DiscussionTopicEntity.js
@@ -69,6 +69,24 @@ export class DiscussionTopicEntity extends Entity {
 		return fields && fields.value;
 	}
 
+	participationSelection() {
+		const options = this.participationOptions();
+		const selected = options?.find(option => option?.selected);
+		return selected?.value;
+	}
+
+	/**
+	 * @returns {object[]} Post rating options of discussion topic.
+	 */
+	participationOptions() {
+		if (!this.canUpdateRatingType()) return;
+
+		const action = this._entity.getActionByName(Actions.discussions.topic.updateParticipationOption);
+		const fields = action && action.getFieldByName('participationType');
+
+		return fields && fields.value;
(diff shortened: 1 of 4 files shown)


martinwang-d2l:
Realized I needed to make additional changes.

src/activities/discussions/DiscussionTopicEntity.js
```diff
@@ -69,6 +69,27 @@ export class DiscussionTopicEntity extends Entity {
 		return fields && fields.value;
 	}
 
+	/**
+	 * @returns {number} participation option of discussion topic.
+	 */
+	participationSelection() {
+		const options = this.participationOptions();
+		const selected = options?.find(option => option?.selected);
+		return selected?.value;
```
aid848, line 78: Should we have a default selection or is it ok to just have null here?

src/activities/discussions/DiscussionTopicEntity.js
```diff
@@ -108,6 +129,14 @@ export class DiscussionTopicEntity extends Entity {
 		return entity && entity.hasActionByName(Actions.discussions.topic.updateRatingType);
 	}
 
+	/**
```
dustxd, line 132: Can you group the related functions together? It's easier to read that way

src/activities/discussions/DiscussionTopicEntity.js
martinwang-d2l: I think it should be fine - paging doesn't use a default either: https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/activities/ActivityUsageCollectionEntity.js#L75

dustxd (approved):

PR #557, opened 2022-12-02
DE51074: Lessons > discard changes dialog appears after FACE page is saved

GurpinderSandhu:
## Relevant Rally Link
https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F671995940833
## Description
**Note**: The defect mentions having to save first but it also occurs without saving

A recent addition of the ability to set the availability type (access restricted until available, submission restricted until available etc.) on some FACE pages was introduced and needed to be added to the dirty check as a result.

In the case where the value of `this.defaultStartDateType()` is undefined (i.e most of lessons FACE pages) `this.defaultStartDateType()?.toString()` would return `undefined` as type undefined, not as type string. So left (type undefined) would not actually equal right (type string) in the comparison causing the check to resolve as dirty and trigger the discard changes dialog.

Using `String()` will cover the case of having an undefined value.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@brightspace-hmc/siren-sdk",
-  "version": "2.48.1",
+  "version": "2.49.0",
   "type": "module",
   "description": "",
   "main": "index.js",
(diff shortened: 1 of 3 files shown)


ianlaidlaw (approved):
nice catch! lgtm when tests pass

Kiefer-Co-D2L (approved):
LGTM!  Sneaky bug.

PR #564, opened 2022-12-09
US145143 Add displayInCalendar for FACE Quizzing as an interim solution

mostlyfabulous:
[US145143](https://rally1.rallydev.com/#/?detail=/userstory/668227308137&fdp=true): [ui] Edit add to calendar

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -301,6 +301,12 @@ export class ActivityUsageEntity extends Entity {
 			const dateEntity = this._getSubEntityByClass(Classes.availabilityDates.availabilityDates);
 			return dateEntity && dateEntity.properties && dateEntity.properties.DisplayInCalendar;
 		}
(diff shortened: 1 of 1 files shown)


mdulat:
Is there an lms change that goes with this?

src/activities/ActivityUsageEntity.js
```diff
@@ -305,7 +305,8 @@ export class ActivityUsageEntity extends Entity {
 		const datesEntity = this._getSubEntityByClass('dates');
 		if (datesEntity) {
 			const action = datesEntity.getActionByName(Actions.activities.update);
-			return datesEntity && action.getFieldByName('displayInCalendar').value;
+			const displayInCalendar = action.getFieldByName('displayInCalendar');
+			return datesEntity && displayInCalendar && displayInCalendar.value;
```
mdulat, line 309: probably don't need the `datesEntity &&` here since it is within the `if (datesEntity)` block. But probably should check for existence of `action` before getting its field.

PR #565, opened 2022-12-14
feat: US146324 add entities for group and section restrictions

mdulat:
[US146324](https://rally1.rallydev.com/#/?detail=/userstory/672768000413&fdp=true): [ui] display summary tile


src/activities/discussions/TopicGroupSectionRestrictionsEntity.js
```diff
@@ -1,8 +1,75 @@
...
+	isRestrictedTopic() {
+		if (!this._entity) return;
+		return this._entity.hasClass(Classes.discussions.restrictedTopic);
+	}
+	hasNoGroupsOrSections() {
+		if (!this._entity) return;
+		return this._entity.hasClass(Classes.discussions.noGroupsOrSections);
+	}
+	isUnRestricted() {
+		return !this.isRestrictedThreads() && !this.isRestrictedTopic() && !this.hasNoGroupsOrSections();
```
aid848, line 23: Should a topic be considered restricted if the course has no groups or sections?

mdulat: No it shouldn't, good catch, I was thinking about all the different states, but the UI already checks for noGroupsOrSections separately, so it doesn't make sense to include that here. I'll remove it.

martinwang-d2l (approved):

aid848 (approved):

PR #581, opened 2023-01-06
DE51142 Remove auto set graded checkbox from dirty check for quiz entity

dustxd:
### Motivation/Context
auto set graded checkbox is now called "Auto-publish attempt results immediately upon completion":

[We are setting the checkbox to true by default if it's a new quiz in FACE](https://github.com/BrightspaceHypermediaComponents/activities/pull/2058) but since this happens after the initial load of quiz entity, new quiz would always be marked as dirty. The idea here is to move dirty check to the component level because the component has information to whether the quiz is new and if save in place has been clicked yet, so it can accurately determine whether the checkbox should be considered dirty.

Related PR: https://github.com/BrightspaceHypermediaComponents/activities/pull/3450

### Rally Defect
[DE51142](https://rally1.rallydev.com/#/?detail=/defect/672370058493&fdp=true): [New Quiz Creation] > Dirty check broken in 23.01

### DIFF
--- a/src/activities/quizzes/QuizEntity.js
+++ b/src/activities/quizzes/QuizEntity.js
@@ -1036,7 +1036,6 @@ export class QuizEntity extends Entity {
 			[this.password(), quiz.password],
 			[this.notificationEmail(), quiz.notificationEmail],
 			[this.isPreventMovingBackwardsEnabled(), quiz.preventMovingBackwards],
-			[this.isAutoSetGradedEnabled(), quiz.autoSetGraded],
 			[this.isSyncGradebookEnabled(), quiz.syncGradebook],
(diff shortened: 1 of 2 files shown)


mdulat:
Will it recognize the quiz as dirty if you edit an existing quiz and toggle that checkbox?

dustxd:
Yes, I added its own dirty function and equality check here: https://github.com/BrightspaceHypermediaComponents/activities/pull/3450/files#diff-c1373be40678baae8550c911ebc996af36e660be5cebe234851576a9bbee6127R75-R77

dustxd:
But let me know if there's another way to approach this! I created these PRs to illustrate my idea and I don't really know if we've solved this problem before 🤔

mdulat (approved):

PR #582, opened 2023-01-10
feat: US146334 add group restriction actions

mdulat:
[US146334](https://rally1.rallydev.com/#/?detail=/userstory/672777856045&fdp=true): [ui] update restriction type

### DIFF
--- a/src/activities/discussions/GroupSectionRestrictionActionsEntity.js
+++ b/src/activities/discussions/GroupSectionRestrictionActionsEntity.js
@@ -1,4 +1,4 @@
-import { Actions, Classes } from '../../hypermedia-constants.js';
+import { Actions } from '../../hypermedia-constants.js';
 import { Entity } from '../../es6/Entity.js';
 import { performSirenAction } from '../../es6/SirenAction.js';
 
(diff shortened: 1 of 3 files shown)


src/activities/discussions/GroupSectionRestrictionActionsEntity.js
```diff
@@ -0,0 +1,44 @@
+import { Actions } from '../../hypermedia-constants.js';
+import { Entity } from '../../es6/Entity.js';
+import { performSirenAction } from '../../es6/SirenAction.js';
+
+export class GroupSectionRestrictionActionsEntity extends Entity {
+	canSetToRestrictedTopic() {
+		if (!this._entity) return;
+		return !this._entity.hasActionByName(Actions.discussions.groupSectionRestrictions.setToRestrictedTopic);
```
aid848, line 8: Should this be negated if we are checking if the action is available?

mdulat: oh yeah, good catch.

aid848 (approved):

PR #587, opened 2023-01-12
DE45495 > Replacement strings not working for modules

GurpinderSandhu:
## Relevant Rally Link
https://rally1.rallydev.com/#/?detail=/defect/610830681123&fdp=true
## Description
Adding support to retrieve the `raw-description` entity's html to be loaded into the editor. Also want to have dirty check to compare against what is in the db not the replaced version.
## Related PRs
https://github.com/Brightspace/lms/pull/32150

### DIFF
--- a/src/activities/content/ContentModuleEntity.js
+++ b/src/activities/content/ContentModuleEntity.js
@@ -20,6 +20,17 @@ export class ContentModuleEntity extends Entity {
 	}
 
 	/**
+	 * @returns {string|null} Raw description html of the content-module item
+	 */
+	rawDescriptionRichText() {
+		const rawDescriptionSubEntity = ContentHelperFunctions.getRawDescriptionSubEntity(this._entity);
+		if (!rawDescriptionSubEntity) {
+			return null;
+		}
+		return rawDescriptionSubEntity.properties.html || '';
+	}
+
+	/**
 	 * @returns {string|null} Description text of the content-module item
 	 */
 	descriptionText() {
@@ -94,7 +105,7 @@ export class ContentModuleEntity extends Entity {
 	equals(contentModule) {
 		const diffs = [
 			[this.title(), contentModule.title],
-			[this.descriptionRichText(), contentModule.descriptionRichText]
+			[this.rawDescriptionRichText(), contentModule.descriptionRichText]
 		];
 		for (const [left, right] of diffs) {
 			if (left !== right) {
--- a/src/activities/content/ContentModuleEntity.js
+++ b/src/activities/content/ContentModuleEntity.js
@@ -105,7 +105,7 @@ export class ContentModuleEntity extends Entity {
 	equals(contentModule) {
 		const diffs = [
 			[this.title(), contentModule.title],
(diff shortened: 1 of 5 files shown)


src/activities/content/ContentModuleEntity.js
```diff
@@ -94,7 +105,7 @@ export class ContentModuleEntity extends Entity {
 	equals(contentModule) {
 		const diffs = [
 			[this.title(), contentModule.title],
-			[this.descriptionRichText(), contentModule.descriptionRichText]
```
ianlaidlaw, line 97: do we want to get rid of `descriptionRichText` or have access to both?

GurpinderSandhu: Since on load, we are now loading in the raw description into the editor, it will be "different" if we compare it with descriptionRichText. I.e loading a FACE page that contains replace strings then pressing cancel without making changes will prompt the discard changes if we keep that line

ianlaidlaw: ah i see i see, sorry you can ignore my comments i was misunderstanding what this section was doing

ianlaidlaw: that might make sense to keep things in sync. it does look a little confusing here when the diff is comparing different properties

kristian-torres: I'd say this is a nit. If you think about the scope of this entity, it does not know a "raw" or "editted" state for the content module description and it is just representing the data for consumers to use. In the LMS side of things it makes sense to keep the name raw because you are formatting the original description text, serializing it, and returning the response. Keeping `descriptionRichText` on the activity side (imo) makes more sense than raw as at the point it is rendering the description.

test/activities/content/data/TestContentModuleEntity.js
```diff
@@ -39,10 +39,17 @@ export const contentModuleData = {
...
+		{
+			'class': ['richtext', 'description'],
+			'properties': { 'text': 'description text', 'html': '<p>description text</p>' },
+			'rel': []
+		},
+		{
+			'class': ['richtext', 'raw-description'],
+			'properties': { 'text': 'description text', 'html': '<p>description text</p>' },
+			'rel': []
+		}
```
kristian-torres, line 52: As for the comment above, we have serialized both the raw and the useable description so having that  distinction on the activities side makes sense (unless I completely misunderstood the PR)

GurpinderSandhu: I was unsure if the description entity is only consumed by us, if it is used elsewhere I did not want to change it. Also, the string replacement [occurs here](https://github.com/Brightspace/lms/blob/6c6462fec92e911f26701f1bb6eec3238945c8bd/lp/framework/core/D2L/LP/TextProcessing/Rendering/Renderers/UserEnteredHtmlHtmlRenderer.cs#L53). There are a bunch of text renderers and it looks like this is the one that is used by FACE to populate the editor.

ianlaidlaw (approved):

Kiefer-Co-D2L (approved):

PR #592, opened 2023-01-30
feat: add new entities for restricted topic type

wkerrd2l:
Implement the group list in the Group and Section Restrictions dialog for restricted-topic types:
[US146337](https://rally1.rallydev.com/#/?detail=/userstory/672777987809&fdp=true)

### DIFF
--- a/src/activities/discussions/GroupEntity.js
+++ b/src/activities/discussions/GroupEntity.js
@@ -1,0 +1,10 @@
+import { Entity } from '../../es6/Entity.js';
+
+export class GroupEntity extends Entity {
+
+	name() {
+		const subEntity = this._entity && this._entity.getSubEntityByClass('name');
+		return subEntity && subEntity.properties && subEntity.properties.name;
+	}
+
+}
--- a/src/activities/discussions/GroupSectionRestrictionActionsEntity.js
+++ b/src/activities/discussions/GroupSectionRestrictionActionsEntity.js
@@ -53,4 +53,30 @@ export class GroupSectionRestrictionActionsEntity extends Entity {
 			return new RestrictedTopicCollectionEntity(entity);
 		});
 	}
+	canToggleGroupsRestrictedTopic() {
+		if (!this._entity) return null;
+		const subEntity = this._entity.getSubEntityByClass('restricted-topic');
+		return subEntity && subEntity.hasActionByName(Actions.discussions.groupSectionRestrictions.toggleGroupsRestrictedTopic);
+	}
+	_formatToggleGroupsRestrictedTopicAction(toggleGroupIds) {
+		if (!toggleGroupIds) return;
+		if (!this.canToggleGroupsRestrictedTopic()) return;
+		const subEntity = this._entity.getSubEntityByClass('restricted-topic');
+		const action = subEntity.getActionByName(Actions.discussions.groupSectionRestrictions.toggleGroupsRestrictedTopic);
+		const fields = [
+			{ name: 'toggleGroupIds', value: toggleGroupIds },
+		];
+
+		return { action, fields };
+	}
+	async toggleGroupsRestrictedTopic(toggleGroupIds) {
+		if (!toggleGroupIds) return;
+		if (!this.canToggleGroupsRestrictedTopic()) return;
+
+		const sirenAction = this._formatToggleGroupsRestrictedTopicAction(toggleGroupIds);
+		if (!sirenAction) return;
+
+		const { action, fields } = sirenAction;
+		await performSirenAction(this._token, action, fields);
+	}
 }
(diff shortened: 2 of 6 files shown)


src/activities/discussions/GroupEntity.js
```diff
@@ -0,0 +1,10 @@
+import { Entity } from '../../es6/Entity.js';
+
+export class GroupEntity extends Entity {
```
mdulat, line 3: I wonder if we should call this `NamedEntity`, since we're using it for both group and group-category, and it could potentially be used for other named entities that use the same format?

src/activities/discussions/RestrictedTopicCollectionEntity.js
```diff
@@ -0,0 +1,41 @@
...
+	nameHref() {
+		return this._entity && this._entity.entities && this._entity.entities[0] && this._entity.entities[0].href;
+	}
+	onGroupChange(onChange) {
+		if (!this._entity) return;
+
+		const href = this.nameHref();
+		if (!href) return;
+
+		this._subEntity(GroupEntity, href, onChange);
```
mdulat, line 39: What is this function doing?

src/activities/discussions/GroupSectionRestrictionActionsEntity.js
```diff
@@ -41,4 +42,41 @@ export class GroupSectionRestrictionActionsEntity extends Entity {
...
+		if (!subEntity) {
+			return null;
+		}
+		return (subEntity.getSubEntitiesByRel('item') || []).map(entity => {
+			return new RestrictedTopicCollectionEntity(entity);
+		});
+	}
+	canToggleGroupsRestrictedTopic() {
+		if (!this._entity) return null;
+		const subEntity = this._entity.getSubEntityByClass('restricted-topic');
```
mdulat, line 58: Let's add this class name into hypermedia-constants as well

src/activities/discussions/GroupSectionRestrictionActionsEntity.js
```diff
@@ -41,4 +42,41 @@ export class GroupSectionRestrictionActionsEntity extends Entity {
...
+		return subEntity && subEntity.hasActionByName(Actions.discussions.groupSectionRestrictions.toggleGroupsRestrictedTopic);
+	}
+	_formatToggleGroupsRestrictedTopicAction(toggleGroupIds) {
+		if (!toggleGroupIds) return;
+		if (!this.canToggleGroupsRestrictedTopic()) return;
+		const subEntity = this._entity.getSubEntityByClass('restricted-topic');
+		const action = subEntity.getActionByName(Actions.discussions.groupSectionRestrictions.toggleGroupsRestrictedTopic);
+		const fields = [
+			{ name: 'toggleGroupIds', value: toggleGroupIds },
```
mdulat, line 67: we'll want to add multiple fields with name `toggleGroupIds`, one for each id to toggle, instead of passing a single string for these ids. Note in [the controller ](https://github.com/Brightspace/lms/blob/master/le/discussions/D2L.LE.Discussions/Extensibility/HM/Discussions/Controllers/DiscussionsController.GroupRestrictions.cs#L47) this param is an array instead of a comma-separated string. These multiple fields will be mapped to this array.

src/activities/discussions/RestrictedTopicCollectionEntity.js
wkerrd2l: I will remove. I think it was from when I was trying to figure out a good way to fetch the Group/Name Entity

src/activities/discussions/RestrictedTopicEntity.js
```diff
@@ -0,0 +1,47 @@
...
+	}
+	async getName() {
+		if (!this._entity) {
+			return null;
+		}
+
+		const action = this.nameHref();
+
+		const fields = [];
+		const returnedEntity = await performSirenAction(this._token, action, fields, false, true);
```
mdulat, line 42: Why do we bypassCache on this call?

src/activities/discussions/RestrictedTopicEntity.js
```diff
@@ -0,0 +1,47 @@
...
+	}
+	nameHref() {
+		return this._entity && this._entity.entities && this._entity.entities[0] && this._entity.entities[0].href;
+	}
+	async getName() {
+		if (!this._entity) {
+			return null;
+		}
+
+		const action = this.nameHref();
```
mdulat, line 39: The action is just a url?

mdulat: I think we can just "fetch" the name from activities, we don't need to hack it into an action here.

wkerrd2l: Yep I think I had this from before when I was trying stuff. I will remove, it not being used with the current implementation.

src/activities/discussions/RestrictedTopicEntity.js
wkerrd2l: Woops, I was playing around with the params and seeing what was going on. This function can be removed.

mdulat (approved):
Nice!

PR #608, opened 2023-03-06
Remove SauceLabs

dlockhart:
We're moving away from using SauceLabs for cross-browser testing and towards Playwright. For more on the reasoning, [see this blog post](https://blog.d2l.dev/2022/03/22/Playwright/). TD;DR: it's cheaper, faster and less flaky.

This PR switches the repo from SauceLabs to Playwright, as well as makes a few consistency tweaks to align it with our current UI repo template.


.eslintrc.json
```diff
@@ -3,8 +3,5 @@
   "rules": {
     "no-prototype-builtins": 0,
     "sort-class-members/sort-class-members": 0
-  },
-  "globals": {
-    "fastdom": false
```
dlockhart, line 8: Wasn't used.

dbatiste (approved):

PR #611, opened 2023-03-07
DE51385 add support for name field in submission views

martinwang-d2l:
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fdefect%2F673800671099

Related LMS PR: https://github.com/Brightspace/lms/pull/33944. 

- From LMS PR: Following discussions with Joseph, it was decided that we will save the additional view name with respect to the users current lang/locale at the time of save. After switching locales, unless we save again, it will remain in the previous locale's translation. For now, we're sticking to this implementation.
- 
- In this PR, added support for the name field for individual submission views, and added test cases.

### DIFF
--- a/src/activities/quizzes/submissionViews/QuizSubmissionViewEntity.js
+++ b/src/activities/quizzes/submissionViews/QuizSubmissionViewEntity.js
@@ -158,6 +158,35 @@ export class QuizSubmissionViewEntity extends Entity {
 		return this._entity && this._entity.hasClass(Classes.quizzes.submissionView.showStatsScoreDistribution);
 	}
 
+	/** NAME SUB-ENTITY */
+
+	_nameSubEntity() {
+		return this._entity && this._entity.getSubEntityByClass(Classes.quizzes.submissionView.viewName.viewName);
+	}
+
+	async setName(value) {
+		const subEntity = this._nameSubEntity();
+		const action = subEntity.getActionByName(Actions.quizzes.submissionView.setName);
+		const fields = [
+			{ name: 'subviewName', value }
+		];
+
+		const returnedEntity = await performSirenAction(this._token, action, fields);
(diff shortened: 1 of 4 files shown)


src/activities/quizzes/submissionViews/QuizSubmissionViewEntity.js
```diff
@@ -158,6 +158,35 @@ export class QuizSubmissionViewEntity extends Entity {
...
 
+	/** NAME SUB-ENTITY */
+
+	_nameSubEntity() {
+		return this._entity && this._entity.getSubEntityByClass(Classes.quizzes.submissionView.viewName.viewName);
+	}
+
+	async setName(value) {
+		const subEntity = this._nameSubEntity();
+		const action = subEntity.getActionByName(Actions.quizzes.submissionView.setName);
```
mdulat, line 169: We should probably check that action exists before trying to execute it.

mdulat: Also that subEntity exist...

src/activities/quizzes/submissionViews/QuizSubmissionViewEntity.js
```diff
@@ -158,6 +158,35 @@ export class QuizSubmissionViewEntity extends Entity {
...
+
+		const returnedEntity = await performSirenAction(this._token, action, fields);
+		if (!returnedEntity) return;
+		return new QuizSubmissionViewEntity(returnedEntity, this._token);
+	}
+
+	viewName() {
+		const subEntity = this._nameSubEntity();
+		if (!subEntity) return;
+		return subEntity.properties.name;
```
mdulat, line 182: Good to be defensive, i.e. `return subEntity.properties?.name`

mdulat (approved):

PR #618, opened 2023-03-24
feat: dedupe a few more mixins

devpow112:
Few more not in the `mixins` folder.


dlockhart (approved):
Deduping all the things!

PR #623, opened 2023-04-17
Add _error property to EntityMixinLit

rzveibild2l:
- [US148957](https://rally1.rallydev.com/#/255540250444d/iterationstatus?detail=%2Fuserstory%2F685985340191&fdp=true&view=607e378b-5b78-43dc-a109-b9177724b734)
- The purpose of this PR is to set an error property when an entity cannot be fetched with `EntityMixinLit`
- This will make it possible to catch the case where a user navigates to a route with an invalid org unit, to be redirected to another page

### DIFF
--- a/src/es6/EntityFactory.js
+++ b/src/es6/EntityFactory.js
@@ -67,7 +67,7 @@ export function entityFactory(entityType, href, token, onChange, entity) {
 	const entityListener = new EntityListener();
 	const onChangeWrapped = (entity, error) => {
 		if (entity) {
-			onChange(new entityType(entity, token, entityListener), null);
+			onChange(new entityType(entity, token, entityListener));
 		} else {
 			onChange(null, error);
 		}
--- a/src/es6/EntityFactory.js
+++ b/src/es6/EntityFactory.js
@@ -67,7 +67,7 @@ export function entityFactory(entityType, href, token, onChange, entity) {
 	const entityListener = new EntityListener();
 	const onChangeWrapped = (entity, error) => {
 		if (entity) {
-			onChange(new entityType(entity, token, entityListener));
+			onChange(new entityType(entity, token, entityListener), null);
 		} else {
 			onChange(null, error);
 		}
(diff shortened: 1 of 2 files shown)


src/mixin/entity-mixin-lit.js
```diff
@@ -65,8 +69,9 @@ const InternalEntityMixinLit = superclass => class extends superclass {
 			const pendingEvent = new AsyncStateEvent(pendingPromise);
 			this.dispatchEvent(pendingEvent);
 
-			entityFactory(this._entityType, this.href, this.token, entity => {
+			entityFactory(this._entityType, this.href, this.token, (entity, error) => {
 				this._entity = entity;
+				this._error = error;
```
rzveibild2l, line 74: To avoid any conflicts, I confirmed with [search.d2l](https://search.d2l.dev/search?full=EntityMixinLit&defs=&refs=&path=&hist=&type=&xrd=&nn=8&searchall=true) that no files which use `EntityMixinLit` make use of an `_error` property.

mdgbayly:
I don't really have any constructive comments. :grimacing: 
We didn't end up using this mixin in FACE components as we transitioned to MobX.
So seems reasonable to me. I assume you will use an updated handler on the _error property or something to catch when an async error occurs?

src/es6/EntityFactory.js
```diff
@@ -67,7 +67,7 @@ export function entityFactory(entityType, href, token, onChange, entity) {
 	const entityListener = new EntityListener();
 	const onChangeWrapped = (entity, error) => {
 		if (entity) {
-			onChange(new entityType(entity, token, entityListener));
+			onChange(new entityType(entity, token, entityListener), null);
```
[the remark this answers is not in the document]
bearfriend: I'm personally not a fan of sending explicit nullish values around that don't serve a real purpose. The lack of sending a value here would indicate the same intention to me, but without creating/initializing something unnecessarily. On the other end, if I received `null` I would think that there was an error with a bad value, whereas if the `error` argument didn't exist at all I would expect success and `entity` to exist.

src/mixin/entity-mixin-lit.js
bearfriend: Even so, it may be too generic and conflict in the future. Maybe `_entityError`?

bearfriend (approved):

zwilson2 (approved):

JeremyWinkler (approved):

rzveibild2l:
@svanherk Just want to confirm that you have no objections, and we are good to merge.

svanherk (approved):

PR #636, opened 2023-05-25
DE53465 - [FACE][Assignments] - Switching submission types in FACE errors

MyraLuetke:
Rally: [DE53465](https://rally1.rallydev.com/#/?detail=/defect/701880762135&fdp=true)

Just a note that any video in the comments depicts bad behaviour that is being fixed by the code it is linked to!


src/activities/assignments/AssignmentEntity.js
```diff
@@ -965,6 +964,9 @@ export class AssignmentEntity extends Entity {
 		if (assignment.hasOwnProperty('allowTextSubmission')) {
 			diffs.push([this.allowTextSubmission(), assignment.allowTextSubmission]);
 		}
+		if (assignment.hasOwnProperty('allowableFileType')) {
+			diffs.push([this.allowableFileTypeValue(), assignment.allowableFileType]);
+		}
```
MyraLuetke, line 969: Only compare allowableFileType between saved entity and state now if the state that is being compared contains the property since with [this change](https://github.com/BrightspaceHypermediaComponents/activities/pull/3933/files#r1205913773) we will not always provide the allowableFileType

jeroach (approved):

JamieTran (approved):

PR #648, opened 2023-07-27
US154016 add action name

mdulat:
[US154016](https://rally1.rallydev.com/#/?detail=/userstory/704807766953&fdp=true): [ui] Render add content dialog

Add action name

### DIFF
--- a/src/hypermedia-constants.js
+++ b/src/hypermedia-constants.js
@@ -665,6 +665,7 @@ export const Actions = {
 		},
 		activityUsageCollection: {
 			setCollectionPaging: 'set-collection-paging',
+			startAddNewActivity: 'start-add-new-activity',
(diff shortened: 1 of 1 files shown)


src/hypermedia-constants.js
```diff
@@ -665,6 +665,7 @@ export const Actions = {
 		},
 		activityUsageCollection: {
 			setCollectionPaging: 'set-collection-paging',
+			startAddNewActivity: 'start-add-new-activity',
```
stevenao, line 668: There is also `d2l-hypermedia-constants` lib. So, I am not sure which one is the main repo. Maybe we need to update both which I hope we standardize it to just 1.

stevenao (approved):

PR #650, opened 2023-08-03
US147082 cleanup Audio Video Editing

shaz3696:
Relevant Rally Links
[US147082](https://rally1.rallydev.com/#/668026692137d/teamboard?detail=%2Fuserstory%2F675232370835)
Description
Acceptance Criteria:

Clean-up code to stop looking for the feature flag and only use the TRUE behaviour

Remove the FALSE behaviour

Remove the Feature Flag

Flag in question:
https://github.com/Brightspace/lms-launch-darkly-flags/blob/master/features/activities/face/content/content-audio-video-advanced-editing.feature.json

Goal/Solution

### DIFF
--- a/test/activities/content/data/TestContentServiceFileEntity.js
+++ b/test/activities/content/data/TestContentServiceFileEntity.js
@@ -41,7 +41,6 @@ export const contentServiceFileData = {
 		'title': 'Test Content Service File Title',
 		'url': 'https://phoenix-is-the-best.com',
 		'embedMedia': 'true',
-		'isAdvancedEditingAvailable': 'true',
 		'isAdvancedEditingEnabled': 'true',
 		'isContentServiceResource': 'true',
 		'contentServiceName': 'content-service-name',
--- a/src/activities/content/ContentServiceFileEntity.js
+++ b/src/activities/content/ContentServiceFileEntity.js
@@ -35,20 +35,6 @@ export class ContentServiceFileEntity extends ContentFileEntity {
 	}
 
 	/**
-	 * @returns {boolean|undefined} Determines whether advanced editing is available for this object type.
-	 */
-	isAdvancedEditingAvailable() {
-		return this._entity && this._entity.properties && this._entity.properties.isAdvancedEditingAvailable;
-	}
-
-	/**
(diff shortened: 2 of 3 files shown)


Letharrick (approved):

test/activities/content/data/TestContentServiceFileEntity.js
```diff
@@ -41,7 +41,6 @@ export const contentServiceFileData = {
 		'title': 'Test Content Service File Title',
 		'url': 'https://phoenix-is-the-best.com',
 		'embedMedia': 'true',
-		'isAdvancedEditingAvailable': 'true',
 		'isAdvancedEditingEnabled': 'true',
```
Kiefer-Co-D2L, line 44: Since this field is also being removed from the siren entity, should it also be removed from this test object?

Kiefer-Co-D2L (changes requested):
I would review the rally ticket and reach out to its creators for more clarification.

The rally ticket only references `content-audio-video-advanced-editing`.  This flag is used when computing `isAdvancedEditingEnabled`, but it is not the only factor in computing that property.

Based on my understanding of the ticket, you should be removing references to the flag when computing `isAdvancedEditingAvailable` in the LMS, which your LMS PR does, but we don't need to remove `isAdvancedEditingAvailable` or `isAdvancedEditingEnabled` from siren or activities, as these fields still have meaning without the LaunchDarkly flag.

shaz3696:
> I would review the rally ticket and reach out to its creators for more clarification.
> 
> The rally ticket only references `content-audio-video-advanced-editing`. This flag is used when computing `isAdvancedEditingEnabled`, but it is not the only factor in computing that property.
> 
> Based on my understanding of the ticket, you should be removing references to the flag when computing `isAdvancedEditingAvailable` in the LMS, which your LMS PR does, but we don't need to remove `isAdvancedEditingAvailable` or `isAdvancedEditingEnabled` from siren or activities, as these fields still have meaning without the LaunchDarkly flag.

Can you please confirm on this

rchaudhary-d2l:
As per suggestions by Gurpinder and Kiefer, this PR is not needed so closing it.

PR #661, opened 2023-09-19
add initial siren entities for instanceCollections and collectedObjects

patel-kajal:
[US157927](https://rally1.rallydev.com/#/110294864140d/custom/110851118712?detail=%2Fuserstory%2F723358925591)

Add basic siren entities for instance collections and collected objects.
Only basic initial properties added for now based on this PR:  https://github.com/Brightspace/lms/pull/41116


src/hypermedia-constants.js
```diff
@@ -56,6 +56,7 @@ export const Rels = {
 		myOrganizationActivities: 'https://activities.api.brightspace.com/rels/my-organization-activities',
 		myOrganizationActivitiesEmpty: 'https://activities.api.brightspace.com/rels/my-organization-activities#empty',
 		activityCollection: 'https://activities.api.brightspace.com/rels/activity-collection',
+		instanceCollection: 'https://activities.api.brightspace.com/rels//InstanceCollection/',
```
kieranderson, line 59: `//` double slash intentional? 

Moving forward we should probably change this to `instance-collection` to match everything else

kieranderson (approved):

src/hypermedia-constants.js
patel-kajal: no it was not intentional, thanks for catching that

src/activities/instanceCollections/CollectedObject.js
```diff
@@ -0,0 +1,22 @@
+import { Entity } from '../../es6/Entity.js';
+
+export class CollectedObjectEntity extends Entity {
+	getSortOrder() {
+		if (!this._entity) {
+			return false;
+		}
+		return this._entity.properties?.SortOrder;
+	}
+	getCompletion() {
```
pureooze, line 10: Is this a "completion type"? Maybe thats a better name? I wasnt sure what completion meant and had to look at the entity below.

adkulas (approved):

src/activities/instanceCollections/CollectedObject.js
patel-kajal: right! Okay, I am going to note this comment, we are still going to be adding a ton here , so we can update this

patel-kajal: I was also debating if we even need to put "get" , maybe just "completionType" is enough? same can be applied to the other functions

PR #663, opened 2023-09-26
Kapatel/usa/us157927 create siren entity collections

patel-kajal:
[US157927](https://rally1.rallydev.com/#/110294864140d/custom/110851118712?detail=%2Fuserstory%2F723358925591)

follow up pr fixing method names.

Related PRs:
1. https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/661
2. https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/663


src/activities/instanceCollections/CollectedObject.js
```diff
@@ -1,13 +1,13 @@
 import { Entity } from '../../es6/Entity.js';
 
 export class CollectedObjectEntity extends Entity {
-	getSortOrder() {
+	sortOrder() {
```
pureooze, line 4: Hmm I personally like having the action verb at the start... 👀

pureooze, line 4: Hmm I personally like having the action verb at the start... 👀

patel-kajal: Thats what I usually prefer for methods, but I was looking at some of the other files and they do it like this.

For example: https://search.d2l.dev/xref/BrightspaceHypermediaComponents/siren-sdk/src/activities/ActivityEvaluationStatusEntity.js?r=6d5e120f#28

What are your thoughts? 🤔

pureooze: Ah you know what that is true. As long as we are dealing with properties, I think this is fine to be consistent 👍

pureooze (approved):

PR #676, opened 2023-11-15
Add paging to audit table entity

JayArora1997:
[US154265](https://rally1.rallydev.com/#/?detail=/userstory/705253427401&fdp=true): Course Merge - Audit Log UI- Paging

Related PR(s):
* https://github.com/Brightspace/ipsis-sis-course-merge/pull/152

PR #679, opened 2023-11-22
fix load more count

JayArora1997:
[US154265](https://rally1.rallydev.com/#/?detail=/userstory/705253427401&fdp=true): Course Merge - Audit Log UI- Paging

When there are more elements available, we want to show the count as "load 20 more" as we increment the page size by 20. 

Related PR(s):
* https://github.com/Brightspace/ipsis-sis-course-merge/pull/152
* https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/676

### DIFF
--- a/src/ipsis/course-merge/BaseCollectionEntity.js
+++ b/src/ipsis/course-merge/BaseCollectionEntity.js
@@ -47,13 +47,14 @@ export class BaseCollectionEntity extends Entity {
 	}
 
 	loadMorePageSize(collectionGetter) {
+		const pageSize = 20;
 		const totalCount = this.totalCount() ?? 0;
 		const collectionLength = collectionGetter()?.length ?? 0;
 		// if pageSize is larger than the number remaining items, return the number of remaining items to be loaded
-		if (totalCount < collectionLength + this.pageSize()) {
+		if (totalCount < collectionLength + pageSize) {
 			return totalCount - collectionLength;
 		}
-		return this.pageSize();
+		return pageSize;
 	}
 }
 


src/ipsis/course-merge/BaseCollectionEntity.js
```diff
@@ -47,13 +47,14 @@ export class BaseCollectionEntity extends Entity {
 	}
 
 	loadMorePageSize(collectionGetter) {
+		const pageSize = 20;
```
kmcdaniel-d2l, line 50: Why not set the pageSize() property to have a default of 20. Assuming this is for when pageinfo does not have a page size

zwilson2: This also breaks the comment just above the if statement. If there are only 3 more items to load, it would display "load 3 more items"

zwilson2 (commented):
See Kevin's comment

src/ipsis/course-merge/BaseCollectionEntity.js
JayArora1997: Testing it again for these cases, Initially it was like this and I changed it to use entity.pageSize thinking the same in my last PR.

JayArora1997:
Dev Testing:

Before:
image

After:
image

src/ipsis/course-merge/BaseCollectionEntity.js
kmcdaniel-d2l: I guess thats fine, I suppose this could revert back to using pageSize(), if we do that refactor to just get next 20 rather than {current} + 20. (idk the status of that)

kmcdaniel-d2l (approved):

zwilson2 (approved):

src/ipsis/course-merge/BaseCollectionEntity.js
JayArora1997: yes, that change would work when paging Api is updated to always fetch the next 20 records instead of doubling the pageSize on every click. I added personal notes to refactor this when that is done!

ckthiessen (approved):

PR #692, opened 2024-01-16
DE56403: remove old "disableRightClick" code

alexli01:
https://rally1.rallydev.com/#/?detail=/defect/734509007647&fdp=true


aid848 (approved):
Makes sense. For reference it was removed from the activities side here: https://github.com/BrightspaceHypermediaComponents/activities/pull/4197

PR #693, opened 2024-01-23
DE56374 > Manual Completion Assignments don't show the manual complete button

GurpinderSandhu:
## Rally Link
https://rally1.rallydev.com/#/?detail=/defect/734336124523&fdp=true
## Description
The manual completion button was being used by content completion and now it is now being used by assigned activities as well, unfortunately in the backend one uses the term `completed` and one uses `complete` , this captures both scenarios without making a change in the backend


Kiefer-Co-D2L (approved):
LGTM!

PR #699, opened 2024-03-01
feat: [IDCS-31] adding description equality check in contentfileentity

venkata-ramana-d2l:
US : https://desire2learn.atlassian.net/browse/IDCS-31

Comparing changes in Description field to display cancel modal.

image

### DIFF
--- a/src/activities/content/ContentFileEntity.js
+++ b/src/activities/content/ContentFileEntity.js
@@ -184,7 +184,8 @@ export class ContentFileEntity extends ContentWorkingCopyEntity {
 	equals(contentFile) {
 		const diffs = [
 			[this.title(), contentFile.title],
-			[this.getFileHref(), contentFile.fileHref]
+			[this.getFileHref(), contentFile.fileHref],
+			[this.descriptionRichText(), contentFile.descriptionRichText]
 		];
 		for (const [left, right] of diffs) {
 			if (left !== right) {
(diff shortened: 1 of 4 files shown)


src/activities/content/ContentFileEntity.js
```diff
@@ -184,7 +184,8 @@ export class ContentFileEntity extends ContentWorkingCopyEntity {
 	equals(contentFile) {
 		const diffs = [
 			[this.title(), contentFile.title],
-			[this.getFileHref(), contentFile.fileHref]
+			[this.getFileHref(), contentFile.fileHref],
+			...(contentFile.fileType !== FILE_TYPES.html ? [[this.descriptionRichText(), contentFile.descriptionRichText]] : [])
```
brissons, line 188: nit: is checking for the html file type necessary?

brissons (commented):
Just one question, but overall lgtm!

I would recommend adding the [Jira ticket link](https://desire2learn.atlassian.net/browse/IDCS-31) to your PR descriptions, since they link up in the Jira ticket sidebar (making development tracking easier):

image

src/activities/content/ContentFileEntity.js
venkata-ramana-d2l: yes, `ContentFileEntity` is a parent class for all the uploaded content files and the equals method is common for all of them. Among those, for HTML file type, we do not have the `descriptionRichText` field. Hence, put a condition to check for that field only in case of non html file types.

brissons (approved):

PR #707, opened 2024-04-03
feat: fed-2590 add aiInspired param on assignment update

mdulat:
https://desire2learn.atlassian.net/browse/FED-2590

### DIFF
--- a/src/activities/assignments/AssignmentEntity.js
+++ b/src/activities/assignments/AssignmentEntity.js
@@ -2,6 +2,9 @@ import { Actions, Classes, Rels } from '../../hypermedia-constants.js';
 import { performSirenAction, performSirenActions } from '../../es6/SirenAction.js';
 import { Entity } from '../../es6/Entity.js';
 
+const HUMAN_GENERATED = 0;
+const AI_INSPIRED = 3;
+
 const actions = {
 	delete: 'delete-folder',
 	cancel: 'cancel-folder'
@@ -266,6 +269,10 @@ export class AssignmentEntity extends Entity {
 		return subEntity.properties.informationText;
 	}
 
+	isAiInspired() {
+		return this._entity && this._entity.hasClass(Classes.assignments.aiInspired);
+	}
+
 	/**
 	 * @returns {bool} If the assignment type is set to individual assignment
 	 */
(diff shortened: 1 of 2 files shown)


src/activities/assignments/AssignmentEntity.js
```diff
@@ -458,6 +465,25 @@ export class AssignmentEntity extends Entity {
...
 
+	/**
+	 * @summary Formats action and fields if assignment has been ai inspired
+	 * @param {object} isAiInspired whether this assignment has been ai inspired
+	 */
+	_formatUpdateAiInspiredAction(isAiInspired) {
+		if (typeof isAiInspired === 'undefined') return;
+		if (!this._hasAiInspiredChanged(isAiInspired)) return;
+
+		const action = this._entity.getActionByName(Actions.assignments.updateName);
```
brunod2l, line 476: I'm confused with this `Actions.assignments.updateName`. Shouldn't this be an ai inspired action?

src/activities/assignments/AssignmentEntity.js
```diff
@@ -458,6 +465,25 @@ export class AssignmentEntity extends Entity {
 		return this._entity && this._entity.hasActionByName(Actions.assignments.updateCustomAllowableFileType);
 	}
 
+	/**
+	 * @summary Formats action and fields if assignment has been ai inspired
+	 * @param {object} isAiInspired whether this assignment has been ai inspired
```
brunod2l, line 470: `@returns` is missing

src/activities/assignments/AssignmentEntity.js
mdulat: yeah, they all have the same endpoint so I'm re-using this action's url instead of adding another one. Our siren-sdk doesn't actually look at the fields in the action, and hard-codes them anyway.

src/activities/assignments/AssignmentEntity.js
mdulat: it's missing in all of them :(

src/activities/assignments/AssignmentEntity.js
brunod2l: ah ok, thanks for explaining it

brunod2l (approved):

PR #716, opened 2024-05-07
NIM-1824 - support immediate flag in add and delete methods for attachments

TristanDevos:
https://desire2learn.atlassian.net/browse/NIM-1824

This allows us to save the attachments immediately if the user on the FACE page chooses to 'Save' and not 'Save and Close'. In the former case, we want the attachments to save/delete immediately so that we can render an accurate representation of what attachments are saved/deleted after the Save occurs.

Dependent PR: https://github.com/BrightspaceHypermediaComponents/activities/pull/4597


jeroach (approved):
lgtm

MyraLuetke (approved):

PR #737, opened 2024-08-19
EP-1112 scorm description properties and set action

ChrisKraljevicD2L:
## JIRA Link
[EP-1112: Introduction of Description Field](https://desire2learn.atlassian.net/browse/EP-1112)

## Description
* SCORM activity entities now have description properties and an accompanying `set` action for richtext description

## Testing
* N/A

## Issues
* Needs LMS PR https://github.com/Brightspace/lms/pull/54764 to work

## Related PRs
1. https://github.com/BrightspaceHypermediaComponents/activities/pull/4803
2. https://github.com/Brightspace/lms/pull/54764

### DIFF
--- a/src/activities/content/ContentScormActivityEntity.js
+++ b/src/activities/content/ContentScormActivityEntity.js
@@ -1,5 +1,6 @@
 import { Actions, Classes } from '../../hypermedia-constants.js';
 import { ContentEntity } from './ContentEntity.js';
+import ContentHelperFunctions from './ContentHelperFunctions.js';
 import { performSirenAction } from '../../es6/SirenAction.js';
 
 /**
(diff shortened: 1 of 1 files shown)


src/activities/content/ContentScormActivityEntity.js
```diff
@@ -21,6 +22,28 @@ export class ContentScormActivityEntity extends ContentEntity {
...
 
+	/**
+	 * @returns {string|null} HTML description of the SCORM activity
+	 */
+	descriptionRichText() {
+		const descriptionSubEntity = ContentHelperFunctions.getDescriptionSubEntity(this._entity);
+		if (!descriptionSubEntity) {
+			return null;
+		}
+		return descriptionSubEntity.properties.html || '';
```
mroberts-d2l, line 33: Is empty string the expected response when there is no html property? It seems like we are returning null above if the subentity cannot be retrieved; does the calling code handle nulls and empty strings differently?

mroberts-d2l (approved):
Other than a potential issue with empty strings (might be the norm, idk), I think this looks good. One note, don't forget to add `fix:` or `feat:` to the beginning of your squash commit. This tells semantic versioning how to version this change.

src/activities/content/ContentScormActivityEntity.js
ChrisKraljevicD2L: This was copy-pasta'd from how the weblinks entity deals with description. I can take a peek into if `null` is treated different from `''` by consumers of these entities, but I doubt it since they're both falsy.

ChrisKraljevicD2L: Drilled down and found this line in [d2l-activity-html-new-editor](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-html-new-editor.js#L47):
`this.value = this.value || '';`
Turns out it'll end up as `''` in the description editor itself whether it's `null` or `''` in the first place lol

mroberts-d2l: Womp womp 🤷‍♀️ . All good then I guess! Thanks for satisfying my curiosity!

kyler-swanson (approved):

PR #738, opened 2024-08-28
Add typescript jsdoc for better IDE integration

awikkerink:
I started looking into adding JSDoc to improve VSCode navigation and tooltips. I've been following https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html. Since this is a commonly used library in our projects, this is a useful place to add typing.

If the IDE doesn't support jsdoc/typescript, there would be no effect, and the extra documentation would be useful.

### DIFF
--- a/src/es6/EntityFactory.js
+++ b/src/es6/EntityFactory.js
@@ -1,7 +1,13 @@
 import 'd2l-polymer-siren-behaviors/store/entity-store.js';
 
+/**
+ * @typedef {import('siren-parser').Entity} ParsedEntity
+ * @typedef {import('siren-parser').Link} SirenLink
+ * @typedef {import('./EntitySirenProperties').EntitySirenProperties} EntitySirenProperties
+ */
+
 /** Allows one to manage the event store listeners. Makes it easy to update, add and remove a listener for the entity store. */
-class EntityListener {
+export class EntityListener {
 	constructor() {
 		this._href;
 		this._token;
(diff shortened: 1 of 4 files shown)


src/es6/EntityFactory.js
```diff
@@ -1,7 +1,13 @@
...
 
+/**
+ * @typedef {import('siren-parser').Entity} ParsedEntity
+ * @typedef {import('siren-parser').Link} SirenLink
+ * @typedef {import('./EntitySirenProperties').EntitySirenProperties} EntitySirenProperties
+ */
+
 /** Allows one to manage the event store listeners. Makes it easy to update, add and remove a listener for the entity store. */
-class EntityListener {
+export class EntityListener {
```
awikkerink, line 10: Exporting so it can be used in a parameter.

src/es6/SirenAction.js
```diff
@@ -3,6 +3,17 @@ import 'd2l-polymer-siren-behaviors/store/entity-store.js';
 import 'd2l-polymer-siren-behaviors/store/action-queue.js';
 import SirenParse from 'siren-parser';
 
+/**
+ * @typedef {import('siren-parser').Action} Action
+ * @typedef {import('siren-parser').Field} Field
+ * @typedef {import('siren-parser').Entity} ParsedEntity
+ * @typedef {{ name: string; value: string }} FieldOverride
```
ryantmer, line 10: Intuitively it seems like this shouldn't be necessary i.e. should be able to just use `Field` - but I assume something is missing somewhere that prevents that?

ryantmer (approved):
bueno

src/es6/SirenAction.js
awikkerink: Field also includes `hasClass`

ryantmer: shoot, right - the exports of `siren-parser` really should have included both `ActualSirenEntity` and `SirenParserEntity`, ah well (this is directed at myself)

PR #740, opened 2024-08-30
Fix jsdoc for performLazySirenAction, add templates for mixins

awikkerink:
I feel like I'm approaching the 80/20 boundary, so I'm finally going to stop after this.

Maybe...

### DIFF
--- a/src/es6/SirenAction.js
+++ b/src/es6/SirenAction.js
@@ -271,7 +271,7 @@ export const performSirenAction = function(token, action, fields, immediate, byp
 /**
  * @param {string|Function} token
  * @param {() => Action} actionFactory
- * @param {FieldOverride[]} fieldOverrides
+ * @param {{ [key: string]: string }} fieldOverrides
  * @param {boolean} [bypassCache]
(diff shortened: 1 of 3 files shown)


src/es6/SirenAction.js
```diff
@@ -271,7 +271,7 @@ export const performSirenAction = function(token, action, fields, immediate, byp
 /**
  * @param {string|Function} token
  * @param {() => Action} actionFactory
- * @param {FieldOverride[]} fieldOverrides
+ * @param {{ [key: string]: string }} fieldOverrides
```
ryantmer, line 274: Are some places overriding fields with something other than `{ name: 'field-name', value: <some arbitrary value> }`? If so, that is technically [non-Siren](https://github.com/kevinswiber/siren?tab=readme-ov-file#name-1) 🤔

ryantmer (approved):

src/es6/SirenAction.js
awikkerink: It's just this method which takes an object

PR #744, opened 2024-09-05
Fix shouldUpdate

awikkerink:
Change caused unit test failures here: https://github.com/Brightspace/d2l-rubric/pull/3352

### DIFF
--- a/src/mixin/entity-mixin-lit.js
+++ b/src/mixin/entity-mixin-lit.js
@@ -47,12 +47,11 @@ const InternalEntityMixinLit = superclass => class extends superclass {
 	}
 
 	shouldUpdate(changedProperties) {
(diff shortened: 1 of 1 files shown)


mpharoah-d2l (approved):

semibran (approved):
I'm not sure how this would have been causing failures. I guess the example for shouldUpdate in lit docs doesn't extend the superclass either

awikkerink:
> I'm not sure how this would have been causing failures. I guess the example for shouldUpdate in lit docs doesn't extend the superclass either

I'm not sure either. I think since shouldUpdate returns true by default and then the LitElement class checks if the shouldUpdate method has been overridden. Not sure how to deal with shouldUpdate in mixins then...

PR #746, opened 2024-09-12
fix: Fix template and optional parameters

awikkerink:


src/mixin/entity-mixin-lit.js
```diff
@@ -4,7 +4,7 @@ import { dedupeMixin } from '@open-wc/dedupe-mixin';
 
 /**
  * @template {new (...args: any[]) => import('lit').ReactiveElement} S
- * @template [EntityType=import('../es6/EntitySirenProperties.js').EntitySirenProperties]
+ * @template {import('../es6/EntitySirenProperties.js').EntitySirenProperties} EntityType
```
awikkerink, line 7: Fixes the setting of `this._entity` on line 84

devpow112 (approved):

PR #748, opened 2024-09-24
add property for aiHumanOrigin for content modules

alogan519:


aid848 (approved):
LGTM. Nit: let's put the Jira link in the description so we can see the PRs in the story more easily.

alogan519:
> LGTM. Nit: let's put the Jira link in the description so we can see the PRs in the story more easily.

Done and also if you put it in the branch name it seems to pick it up that way as well.

aid848:
> > LGTM. Nit: let's put the Jira link in the description so we can see the PRs in the story more easily.
> 
> Done and also if you put it in the branch name it seems to pick it up that way as well.

Ohh that's cool

PR #755, opened 2024-10-22
[MAJ-1292] LTI - Add AP deep links & attached processors

jguann:
[MAJ-1292](https://desire2learn.atlassian.net/browse/MAJ-1292)

- Add methods to access (2) AP hrefs from `AssignmentEntity`
- Add CollectionEntity and Entity for AP Deep Links
- Add CollectionEntity and Entity for AP Attached Processors
- Add test data and tests for the new Collection/Entities

The payloads follow the HM Contract here: https://desire2learn.atlassian.net/l/cp/k96B4e1F


src/activities/assetProcessor/APAttachedProcessorsCollectionEntity.js
```diff
@@ -0,0 +1,94 @@
+import { Classes } from '../../hypermedia-constants.js';
+import { Entity } from '../../es6/Entity.js';
+
+class APAttachedProcessorEntity extends Entity {
+
+	/**
+	 * @returns {string} The external deployment id of the deployment this processor belongs to.
+	 */
+	externalDeploymentId() {
+		return this._entity && this._entity.properties && this._entity.properties.externalDeploymentId;
```
owenboydd2l, line 10: This is interesting. I see it used in other places and reading it would expect it to return true/false but I'm assuming it sort of front-loads true/false checks and then returns the actual value kinda?

really weird

owenboydd2l (approved):
lgtm!

One step closer!

owenboydd2l (approved):

PR #766, opened 2025-01-10
Change to incremental release to auto release work

mostlyfabulous:
It would be better to use `match-lms-release` but that would need a freeze on BSI to update all the consumers to pick up the new major and versioning formatting.


mdulat (approved):
Yay!

PR #778, opened 2025-06-25
FED-5892 Add new property to check if AI modify content functionality enabled.

jkyon:
Add `isAIContentEnabled` to activityUsageEntity to be used by pages to decide whether to show the "modify with AI" button.

Story Related
https://desire2learn.atlassian.net/browse/FED-5892

requires
https://github.com/Brightspace/lms/pull/69313

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -682,6 +682,13 @@ export class ActivityUsageEntity extends Entity {
 	}
 
 	/**
+	 * @returns {bool} Whether or not the activity usage entity has AI modify content functionality enabled
+	 */
+	isAIContentEnabled() {
+		return this._entity && this._entity.hasClass('ai-content-enabled');
+	}
+
+	/**
 	 * @returns {bool} Whether or not the activity usage entity is draft
 	 */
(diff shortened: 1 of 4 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -681,6 +681,13 @@ export class ActivityUsageEntity extends Entity {
 		return this._entity && this._entity.hasClass('ai-summary-enabled');
 	}
 
+	/**
+	 * @returns {bool} Whether or not the activity usage entity has AI modify content functionality enabled
+	 */
```
aid848, line 686: Is it worth putting anything in the `ActivityUsageEntity.test.js` test file?

src/hypermedia-constants.js
```diff
@@ -400,6 +400,7 @@ export const Classes = {
 		rawDescription: 'raw-description',
 		lastModified: 'lastModified',
 		aiInspired: 'ai-inspired',
+		AIContentEnabled: 'ai-content-enabled'
```
alogan519, line 403: looks like this should be camelcase - aiContentEnabled to match aiInspired

PR #782, opened 2025-07-23
[ MAJ-2143 ] Add Requires EULA to APDeeplinkCollectionEntity

rachellh123:
https://desire2learn.atlassian.net/browse/MAJ-2143

Adding "requires eula" to deeplinking entity to be used when attaching an asset processor


src/activities/assetProcessor/APDeepLinksCollectionEntity.js
```diff
@@ -33,6 +33,10 @@ class APDeepLinkEntity {
 	deepLinkLaunchRoute() {
 		return this._entity && this._entity.properties && this._entity.properties.deepLinkLaunchRoute;
 	}
+
+	requiresEula() {
+		return this._entity && this._entity.properties && (this._entity.properties.requiresEula ?? false);
```
owenboydd2l, line 38: should always include this value, but extra careful check if things get merged out-of-order

PR #786, opened 2025-10-01
Adding check for is resusable activity

kieranderson:
[USA-1673 - Createspace Hypermedia refactoring](https://desire2learn.atlassian.net/browse/USA-1673)

### DIFF
--- a/src/activities/ActivityUsageEntity.js
+++ b/src/activities/ActivityUsageEntity.js
@@ -251,6 +251,20 @@ export class ActivityUsageEntity extends Entity {
 	}
 
 	/**
+	 * @returns {bool} Whether or not the activity is reusable
+	 */
+	isReusableActivity() {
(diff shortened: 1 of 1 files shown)


src/activities/ActivityUsageEntity.js
```diff
@@ -250,6 +250,20 @@ export class ActivityUsageEntity extends Entity {
 		activityCollectionHref && this._subEntity(ActivityUsageCollectionEntity, activityCollectionHref, onChange);
 	}
 
+	/**
+	 * @returns {bool} Whether or not the activity is reusable
+	 */
+	isReusableActivity() {
+		return this._entity.hasClass('reusable-activity');
```
adkulas, line 257: Consider:
Putting the string into the classes folder

Could this error if the entity is null?

PR #800, opened 2026-06-15
GAUD-10175 - Update linting and sinon

svanherk:
Noticed the Lit Dependabot PR here and timeboxed getting that updated and updating the linting.

### DIFF
--- a/test/activities/assignments/AssignmentEntity.test.js
+++ b/test/activities/assignments/AssignmentEntity.test.js
@@ -271,24 +271,6 @@ describe('AssignmentEntity', () => {
 			expect(fetchMock.called()).to.be.true;
 		});
 
-		it('saves custom allowable file type', async() => {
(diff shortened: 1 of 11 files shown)


test/activities/assignments/AssignmentEntity.test.js
```diff
@@ -271,24 +271,6 @@ describe('AssignmentEntity', () => {
 			expect(fetchMock.called()).to.be.true;
 		});
 
-		it('saves custom allowable file type', async() => {
```
svanherk, line 274: As far as I can tell, this is the exact same as the test on line 144. But please check me on that 👀

dlockhart: Haha yep looks the same to me... weird.

dlockhart (approved):

